PR #5, opened 2012-12-06
implement grouped animation function.

shans:


web-animation.js
```diff
@@ -1115,6 +1116,36 @@ AnimFunc._createKeyframeFunc = function(property, value, operation) {
 }
 
 /** @constructor */
+var GroupedAnimFunc = function() {
+  GroupedAnimFunc.$super.call(this);
+  this.children = [];
```
dstoc, line 1121: Need a length property, todo?

PR #15, opened 2012-12-12
A couple of minor fixes, and some tests to document the odd behaviour regarding default start times in parent groups

steveblock:


steveblock:
Ready for a look ...

PR #17, opened 2012-12-12
refactor the way we do add, interpolate, fromCssValue, toCssValue.

shans:


web-animation.js
```diff
@@ -1708,20 +1621,121 @@ var fromCssValue = function(property, value) {
...
-      return {r: Number(r[1]), g: Number(r[2]), 
-          b: Number(r[3]), a: Number(r[4])}
-    }
-    return colorDict[value]; 
-  } else {
-    throw 'UnsupportedProperty';
   }
+};
+
+var supportedProperties = new Array();
```
dstoc, line 1627: Not actually an array.

Might be cleaner to define as:

supportedProperties = {
  left: lengthType,
  ...
}

svgProperties = {
  cx: true,
  x: true,
  ...
}

PR #30, opened 2012-12-14
A few more bug fixes

steveblock:


shans:
I think this one won't merge because of the last patch - the iterationTime bug is already fixed in head.

steveblock:
Ah, I see - https://github.com/web-animations/web-animations-js/commit/bd304e461dadca304829a8ddf7b56849b701035f - but I think this fix is incomplete as it's missing the corresponding change to use an inclusive start bound when applying fill modes, and doesn't update the tests.

I've rebased my change.

shans:
I made that change to fix an issue in a different test :)

This is good to go in.

PR #34, opened 2012-12-19
A couple of minor fixes

steveblock:


dstoc:
Is there a test that covers the start time being set correctly?

steveblock:
No, there isn't. parentGroup._addChild() causes updateIterationDuration() to be called on the item. So with the existing code, immediately after updateIterationDuration() is called with the wrong start time, it's called again with the correct time, and there's no visible error.

I ran into this problem while playing with custom time sources, and in this case, the error is visible.

PR #42, opened 2013-01-08
Test from sandy

sandyphan:
A demo with 2 circles put into groups and move in different direction as well as changes colour. Problem happens when 2 squares that translate and rotate are added

### DIFF
--- a/demos/test-circle-square.html
+++ b/demos/test-circle-square.html
@@ -1,0 +1,107 @@
+<!-- 
+Copyright 2012 Google Inc. All Rights Reserved.
+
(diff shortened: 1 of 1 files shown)


steveblock:
Hi Sandy. Did Shane have any more comments for you to address on this? If not, can you squash these commits into 1 commit and push it to your fork (no need for a new pull request). I'll then merge it.

I'll also look into the bug you mention.

sandyphan:
I believe that Shane has no other comments and I have squashed them into 1 commit.

PR #55, opened 2013-01-17
Fix style in tests

steveblock:
This replaces tabs with spaces, and fixes indentation.

We're about to mess with all of the tests when we add the 'attach()' syntax for TimeSources, so this seems like a good time to clean this up.

### DIFF
--- a/demo_example.html
+++ b/demo_example.html
@@ -122,8 +122,7 @@ function first() {
 }
 
 function second() {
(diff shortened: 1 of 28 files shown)


dstoc:
Can you rebase? It doesn't merge cleanly.

steveblock:
done

Steve Block

Google Australia

PR #60, opened 2013-01-17
Fix svg-anim test

dstoc:


shans:
Steve totally fixed this first :)

PR #67, opened 2013-01-23
More fixes in preparation for switching to TimeSources

steveblock:


shans:
Looks good. I think Doug is right about using children.push instead of x = x.concat - good to merge once this is updated.

Issue #65, opened 2013-01-22
ParGroup call causing a maxium call stack error


StephMc:
new Animation(topObjects[i], {left: ["100px", "600px"]}, 2);
new Animation(topObjects[i], {left: ["100px", "600px"]}, 2);
var parentAnimation = new ParGroup(document.animationTimeline.children);

The above code causes an error "Uncaught RangeError: Maximum call stack size exceeded"

steveblock:
Thanks for the bug report Steph. Could you write a test that covers this situation (based on the code you have above) and submit a pull request? I'll try to fix the bug and will include the test when I do.

steveblock:
The problem is that parentGroup tries to add itself as a child, causing an infinite recursion. This happens because when the underlying AnimationGroup is created, it is added to document.animationTimeline.children. When the ParGroup then tries to add all of the specified children from document.animationTimeline.children, it encounters itself. (There's an indexing subtlety which means we only hit this condition when the array of children initially has at least two members)

The fix is to take a copy of the array of children before creating the AnimationGroup, to avoid the effects of it being later modified.

StephMc:
https://github.com/web-animations/web-animations-js/pull/68

steveblock:
Fixed with https://github.com/web-animations/web-animations-js/commit/d10a587effe26a4267b4e1ce4dc29deea9ef02a5

PR #68, opened 2013-01-24
Added test which causes the parGroup problem

StephMc:


steveblock:
Issue https://github.com/web-animations/web-animations-js/issues/65 was fixed with https://github.com/web-animations/web-animations-js/commit/d10a587effe26a4267b4e1ce4dc29deea9ef02a5

PR #69, opened 2013-02-03
Supports chrome browser (v11) on google tv

ragingwind:
i've tested this project on chrome browser on google tv(v11, logitech revue). unfortunately it had not work. so i've changed code about define property. it's just reporting for issue below.

add setter to avoid exception [1] and using defineProperty instead of **defineGetter** for inheritance error [2]

[1]: TypeError: Cannot set property interationTIme of par 0-NaN(0@0) [par 1-undefined (0 @-0.994)[] ] which has only a getter
[2]: can't found super object (ParGroup) if it's using **defineGetter**

### DIFF
--- a/web-animation.js
+++ b/web-animation.js
@@ -147,7 +147,7 @@ var TimedItem = function(timing, startTime, parentGroup) {
   }.bind(this));
   this._startTime = startTime;
   this.currentIteration = null;
-  this._iterationTime = null;
+  this.iterationTime = null;
   this.animationTime = null;
   this._reversing = false;
 
@@ -179,14 +179,6 @@ var TimedItem = function(timing, startTime, parentGroup) {
   this.updateIterationDuration();
   this._pauseStartTime = 0;
 };
-
-TimedItem.prototype.__defineGetter__('iterationTime', function() {
-  return this._iterationTime;
-});
-
-TimedItem.prototype.__defineSetter__('iterationTime', function(time) {
-  this._iterationTime = time;
-});
 
 TimedItem.prototype.__defineGetter__('timeDrift', function() {
   if (this.locallyPaused) {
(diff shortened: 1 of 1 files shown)


ragingwind:
I am sorry that I confused you with previous commit. please refer this comment.
i've removed useless and confues code. i'm trying to solve a type error [1] on chrome of google tv. the chrome raised exception [1]. i just changed declaration code of DEFAULT_GROUP, **defineGetter** to Object.defineProperty.

[1]: TypeError: Cannot set property interationTIme of par 0-NaN(0@0) [par
1-undefined (0 @-0.994)[] ] which has only a getter

steveblock:
I'm afraid I'm still not sure what the problem is. It sounds like you're having problems assigning to document.animationTimeline.iterationTime? When I try in Chrome on Linux, this does not cause an error.

Can you attach a test case? Have you repro'd this in regular Chrome, or just in Chrome for Google TV?

ragingwind:
i also don't know exactly what is problem. i've tested chrome on linux and
mac, i never got any error, it was works very well. i have a exception only
on chrome for google tv. i just would like to report to you this exception.

steveblock:
OK. Can you give me a repro case for the error, or at least explain what you were doing when you observed it?

ragingwind:
i tested with 'test-to-animation.html' in demo and i just logged the exception [1]

[1]:  TypeError: Cannot set property interationTIme of par 0-NaN(0@0) [par  1-undefined (0 @-0.994)[] ] which has only a getter

dstoc:
Things have changed significantly since this issue was logged. I tested that particular test on logitech revue and it seems to work now.

ragingwind:
great! thank :+1:

PR #72, opened 2013-02-04
Extra-asserts with only the menu bars and page structure

StephMc:


StephMc:
I've updated the code with the suggested changes

StephMc:
I've updated the code with the suggested changes

StephMc:
I've updated the code with the suggested changes

PR #76, opened 2013-02-07
Added functionality: Tests can be created via check()

StephMc:


StephMc:
I've added some more functionality to the code

StephMc:
I've updated the code with the suggested changes.

StephMc:
I've updated the code with the suggested changes.

PR #79, opened 2013-02-07
Add content-loading framework for tutorial

sandyphan:


steveblock:
It looks like the commit message is from the last in a series of commits that you've squashed together. Can you update it to reflect what this commit does?

You should follow the file naming style of the existing code, lowercase with dashes. Also, we generally don't abbreviate, other than commonly used cases. So use basic-animations/basic-animation.html etc.

Also, I think tutorial/ can be a top-level directory, alongside test/, rather than inside it. tutorial/ depends on test/, but not the reverse, right?

sandyphan:
Yeah, sorry, I meant loadPage.js

sandyphan:
How do I update the message of the commit for the 2 html files and loadPage.js?

sandyphan:
I have just moved the tutorial folder and placed it on the same level as test/

sandyphan:
I have also renamed the files

sandyphan:
I don't specify a href because href will be generated according to the name of the file using currentSection, as currentSection specify which topic it is in, so if I'm currently in basic-animation, currentSection would return basic-animation and will load basic-animation.html

sandyphan:
Should I push up another commit for change file names? That would make a small change to loadPage.js file as well.

sandyphan:
I have just updated the commits. Would you like me to squash them too?

steveblock:
Yes, let's squash all of the commits together, other than the first commit to add jquery

steveblock:
Ping me when you push an update, as I don't get notifications for new pushes, only new comments.

sandyphan:
I have pushed up new updates, do you want me to squash them again as well?

steveblock:
Looks good. Just squish the last 4 commits together, and edit the first commit to put jquery.js in the right place, then i'll merge it.

sandyphan:
I have rebased the commits into 2 commits, having the first commit to be uploading jquery.js.

sandyphan:
Sorry I might have misunderstood your requirements. For jquery.js to be in the right place do you mean to put it in the main directory outside of the tutorial directory?

PR #83, opened 2013-02-11
Added the animation time viewer

StephMc:


StephMc:
I've updated the code as suggested

PR #84, opened 2013-02-11
Made loading of try-it editor dynamically along with the contents

sandyphan:


steveblock:
I think we need some shuffling around of these commits. There should be ...
- a commit which adds loadEditor(), including nextId() and the animNum div (though see my comments on changing this)
- a second commit which adds the new parallel tutorial
- a third commit which fixes up the style in existing code

Let me know if you need a hand with the git surgery

steveblock:
I've taken this code and moved it to https://github.com/web-animations/web-animations-js/pull/90, with the commits reorganised.

PR #85, opened 2013-02-13
Added testPacket sorting and auto test runner

StephMc:


StephMc:
I've updated the code with the suggestions

PR #88, opened 2013-02-15
Added assert_properties

StephMc:


test/extra-asserts.js
```diff
@@ -245,7 +254,80 @@ function autoTestRunner(){
...
+// Takes in a testRecord
 function assert_properties(test){
+  var object = test.object;
+  var targets = test.targets;
+  var message = test.message;
+  var epsilon = test.epsilon;
+  // Displays the actual current time, not just the time the test
+  // thinks it is (from test.time)
+  var time = parentAnimation.iterationTime === null ?
+      0 : parentAnimation.iterationTime;
```
DmitryBaranovskiy: `var time = Number(parentAnimation.iterationTime);`

test/extra-asserts.js
```diff
@@ -245,7 +254,80 @@ function autoTestRunner(){
...
+
+  // SVGs have to be handled differently to divs
+  var isSVG = (object.nodeName != "DIV");
+
+  // Create an element of the same type as testing so the style can be applied
+  // from the test. This is so the css property (not the -webkit-does-something
+  // tag) can be read.
+  var tempOb = document.createElement(object.nodeName);
```
DmitryBaranovskiy: Shouldn’t it be `createElementNS("http://www.w3.org/2000/svg", object.nodeName);` for SVG?

StephMc:
I've updated the code with the suggestions

StephMc:
I've updated the code with the suggestions

StephMc:
I've updated the code with the suggestions

PR #91, opened 2013-02-19
Clean up some hotspots and simplify sort order calculation

dstoc:


steveblock:
This is incorrect (given our understanding of the current spec) of as it does not take into account iterations.

In any case, it is now moot as we plan to base compositing order on tree order - https://github.com/web-animations/web-animations-js/issues/100

dstoc:
I still need to pull across some of the other fixes.

dstoc:
Ready to merge now.

PR #102, opened 2013-02-22
Use the document load event as the zero point for document time

### LINKED ISSUE #101: Additive animations using Web Animations API issue with getComputedStyle(element).transform and matrix3d
Currently I am trying to "simulate" additive animations using Web Animations API.

As @dstockwell pointed out in this issue https://github.com/web-animations/web-animations-js/issues/14 I am using `getComputedStyle(element)`.

The problem I am facing is that the matrix3d returned by `getComputedStyle(element).transform` probably does not contain enough information on perspective, so if I am using as start keyframe, animations are not the same (issue with Red box)

In this example, the blue box has the desired animation, red box animation is not correct:

https://jsfiddle.net/gibbok/y3mbqtah/

I would like to know:
- If a solution to this problem exists
- If the implementation of additive animations would support `transform` with `perspective`.

Thanks for your time.

```html
<div id="wrapper">
  <div id="target"></div>
</div>

<div id="wrapper2">
  <div id="target2"></div>
</div>
```
```js
var elem = document.getElementById('target');
var style = window.getComputedStyle(elem);

elem.animate([{
  transform: style.transform // using matrix3d returned from getComputedStyle()
}, {
  transform: 'rotateY(179deg) scaleZ(2) translateZ(200px)'
}], {
  duration: 3000
});

var elem2 = document.getElementById('target2');
elem2.animate([{
  transform: 'rotateY(-179deg) scaleZ(2) translateZ(200px)'
}, {
  transform: 'rotateY(179deg) scaleZ(2) translateZ(200px)'
}], {
  duration: 3000
});
```
```css
#wrapper,
#wrapper2 {
  perspective: 1000px;
}

#target,
#target2 {
  position: absolute;
  top: 100px;
  width: 200px;
  height: 300px;
}

#target {
  left: 100px;
  background-color: red;
  transform: rotateY(-179deg) scaleZ(2) translateZ(200px);
}

#target2 {
  left: 400px;
  background-color: blue;
  transform: rotateY(-179deg) scaleZ(2) translateZ(200px);
}
```

steveblock:
Also make sure that the document time is null before this zero point, and add a
test to check that the document time is constant within a JavaScript
callstack.

This fixes https://github.com/web-animations/web-animations-js/issues/101

### DIFF
--- a/web-animation.js
+++ b/web-animation.js
@@ -177,7 +177,7 @@ var Player = function(timedItem, timeline) {
   this._timeline = timeline;
   this._startTime = this._timeline.currentTime;
   this._timeDrift = 0.0;
-  this._pauseTime = null;
+  this._pauseTime = undefined;
 
   this.timedItem.timeDrift = 0.0;
   this._update();
@@ -206,7 +206,7 @@ Object.defineProperty(Player.prototype, 'timedItem', configureDescriptor({
 Object.defineProperty(Player.prototype, 'currentTime', configureDescriptor({
   set: function(currentTime) {
     // This seeks by updating _drift. It does not affect the startTime.
-    if (this._pauseTime === null) {
+    if (!isDefined(this._pauseTime)) {
       this._timeDrift =
           this._timeline.currentTime - this.startTime - currentTime;
     } else {
@@ -215,7 +215,10 @@ Object.defineProperty(Player.prototype, 'currentTime', configureDescriptor({
     maybeRestartAnimation();
   },
   get: function() {
-    return this._pauseTime === null ?
+    if (this._timeline.currentTime === null) {
+      return null;
+    }
+    return !isDefined(this._pauseTime) ?
         this._timeline.currentTime - this._timeDrift - this.startTime :
         this._pauseTime;
   },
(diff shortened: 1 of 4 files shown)


web-animation.js
```diff
@@ -2393,26 +2393,64 @@ var stableSort = function(array, compare) {
...
+  return useHighResolutionTiming ? performance.now() : Date.now();
 };
-var clockZero = clock();
+// Set up the zero times for document time. Document time is relative to the
+// load event.
+var documentTimeZeroAsRafTime = undefined;
+var documentTimeZeroAsClockTime = undefined;
+if (useHighResolutionTiming) {
+  addEventListener('load', function() {
+    // RAF time is relative to the navigationStart event.
```
dstoc: I don't think this is true of webkitRequestAnimationFrame (the comment that was on 2396), can you test in Safari?

steveblock: I thought that comment simply related to whether or not high-performance timing is available? What's important here is the origin used by the RAF time. I'll investigate.

steveblock:
I'm checking some test failures, so please hold off on this one for now ...

steveblock:
Rebased ontp some other fixes, so ready for another look

PR #109, opened 2013-03-05
Added slide functionality & user pause

StephMc:


steveblock:
Nice - just a few small things

StephMc:
I've updated the code with the suggested changes.

steveblock:
Just a couple more things

StephMc:
I've updated the code with the suggested changes.

StephMc:
I've updated the code with the suggested changes.

steveblock:
One minor thing, which you can address later.

PR #116, opened 2013-03-06
Implement Element.animate

dstoc:


steveblock:
Need to add test to test-harness.html

PR #120, opened 2013-03-06
Use ES5 getters and setters as well as __proto__ where available.

arv:
This fixes pure JS issues with IE9+

This cleans up the code to use ES5 getters and setters in the object literals instead of reflection methods.

This also uses **proto** in the object literals (wrapped in createObject for IE support)

### DIFF
--- a/web-animations.js
+++ b/web-animations.js
@@ -16,18 +16,13 @@
  */
 (function() {
 
-var inherits = function(child, parent) {
-  var tmp = function() {};
-  tmp.prototype = parent.prototype;
-  child.$super = parent;
-  child.prototype = new tmp();
-  child.prototype.constructor = child;
-};
-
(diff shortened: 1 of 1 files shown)


steveblock:
Thanks Erik. Just a couple of minor things, but it looks like this will need rebasing anyway.

arv:
configureDescriptor is still used.

Where is the indentation wrong? The comment is not attached to the code

arv:
PTAL

steveblock:
LGTM

Will leave it to Dmitry

DmitryBaranovskiy:
LGTM. :)

steveblock:
It looks like this needs rebasing - if you can do so I'll merge it.

arv:
Rebased

PR #124, opened 2013-03-07
Added the transforms special case and ability to hide flashes

StephMc:


StephMc:
I've updated the code with the suggestions.

PR #129, opened 2013-03-11
Add support for dynamically loading polyfill (after document has loaded).

sorvell:
Adds support for dynamically loading the polyfill by setting up 0 times if the document is already loaded at the time the polyfill script is processed.

This is done by checking document.readyState == 'complete'.


sorvell:
Done. If this weren't a polyfill for a platform feature, I might argue that document zero time just be defined as when animations support is loaded, but I think you guys are right and a warning is the right approach here.

Thanks for the prompt attention

PR #140, opened 2013-03-12
Detect which calc function and transform property should be used

dstoc:


shans:
I'm going to merge despite my comment. If you agree with it, then I'll set up another patch to add the comment as I'm causing extra work :)

dstoc:
OK!

sorvell:
This is awesome. Thanks guys.

PR #154, opened 2013-03-13
Add Timeline.getCurrentPlayers

dstoc:
TestRunner still needs a little work since setting currentTime can cause a Player to no longer be current.


shans:
OK to merge after a rebase

PR #156, opened 2013-03-14
Reduce boilerplate in new test framework

dstoc:


shans:
Looks good. So we still need to extract test results into separate files? Do you want this merged despite that not being done?

dstoc:
It works either way. Not sure it's worth refactoring all these tests since they aren't quite the same as the main test cases.

PR #157, opened 2013-03-14
Switch to the new framework

dstoc:
Not ready to land yet since although the runner will iterate through all the tests, none of them do anything.

Still need to:
- add checks (in separate test-name-checks.js files)
- refactor tests and possibly framework to accomodate tests that do more than set up players before load


dstoc:
Ready now.
- Tests work as before under test-harness.html
- Ported test-bounds.html as manual test example

Issue #162, opened 2013-03-19
Add bounding box testing path to test-generator.html and anim-test-pre.js


shans:
SVG element transforms do not appear in computed style (at least in webKit). We need a way to test these that will work both in the polyfill and in native code. Probably the easiest way is to add bounding box testing.

PR #166, opened 2013-03-19
Added 'ctm' as a valid property type in test-generator and anim-test-pre...

shans:
....

Used 'ctm' to generate checks of test-compositor.html.


shans:
This addresses #162

PR #176, opened 2013-03-25
Stop using Comic Sans

dstoc:


shans:
I dunno, I think this really detracts from the character of the test framework.

Maybe we can use Papyrus instead?

PR #195, opened 2013-04-22
Changes to the test harness

mithro:
Making the test-runner run all the tests in parallel and use XHR requests.

### DIFF
--- a/test/test-runner.html
+++ b/test/test-runner.html
@@ -1,6 +1,8 @@
 <!--
 Copyright 2012 Google Inc. All Rights Reserved.
 
+vim: set ts=2 sw=2 et sts=2 ai:
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


dstoc:
I'm surprised this runs faster given that the tests still run on the same thread.

Is it just that we're able to load the test resources in parallel?

mithro:
I didn't check to see if the XHR change or the parallel tests makes it run faster.

mithro:
Old

```
Ran 3032 tests in 25.978s
OK

real    0m26.737s
```

New

```
Ran 3032 tests in 12.327s
OK

real    0m13.588s
```

Most of the time now is left in starting up / stopping Chrome.

test/test-runner.html
```diff
@@ -19,205 +21,256 @@
...
+  color: green;
+}
 
-var autoAdvance = true;
 
-var currentTest = 0;
-var currentTestStart;
+/* Pass styles */
+tbody.no-tests {
+  color: yellow;
```
dstoc, line 45: Yellow is hard to read, maybe orangered?

dstoc:
Can you add something so that when I run the test-runner manually I can easily know when it's finished?

dstoc:
Need to add back the measured run time for each test.

mithro:
Time measurements make no sense when the tests are run in parallel?

dstoc:
It shows us how long a test takes to run relative to other tests.

mithro:
But a test could be blocked on another test, they are no longer independent
if run in parallel?

PR #197, opened 2013-04-26
5 more tests converted.

mithro:
They are all pretty straight forward.

Tim

### DIFF
--- a/test/testcases/test-compositing-order.html
+++ b/test/testcases/test-compositing-order.html
@@ -42,8 +42,6 @@ var leftToRight = {left: ["100px", "200px"]};
 var rightToLeft = {left: ["200px", "100px"]};
 var noop = {color: "red"};
 
-var dt = document.timeline;
-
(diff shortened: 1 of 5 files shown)


test/testcases/test-compositing-order.html
```diff
@@ -48,20 +31,19 @@
...
+var check = function(actual, expected, description) {
+  console.log('check', actual, expected);
+  test(function() { assert_equals(actual, expected); }, description);
 }
 
 var leftToRight = {left: ["100px", "200px"]};
 var rightToLeft = {left: ["200px", "100px"]};
 var noop = {color: "red"};
 
+var dt = document.timeline;
```
dstoc: Is this needed?

mithro: I guess not, but it seems what the rest of the tests do?

dstoc: Indeed, I'm not sure why. Seems silly to me. You can merge this change as it is or fix that.

PR #199, opened 2013-04-29
A bunch more tests have been converted.

mithro:
Tests remaining to be converted;
- test-change-playback-rate.html - Position dependent on frame rate, need to override callback.
- test-preset-timings.html - Draws on a canvas, no idea how to test.
- test-start-time-iterations.html - Not all the animation stuff is being intercepted, so the moving forward/backward in time isn't working.
- test-step.html - Another one which draws on canvas.


test/testcases/test-unpause.html
```diff
@@ -16,76 +16,52 @@
...
-        "unpausing");
+  check(player.currentTime, 5.0,
+    name + " Player.currentTime should be unaffected by unpausing");
 };
 
 // Test that a Player's currentTime is correct after pausing and unpausing,
 // with a zero startTime.
-addEventListener("load", function() {
-  testPauseUnpause(document.timeline.play(new Animation(
+setTimeout(function() {
```
dstoc: Is there an issue with using the load event? This could be flaky.

mithro: The newer test runner loads all the tests before running them. Would that break this?

mithro: Fixed.

PR #200, opened 2013-05-09
fix for firefox

morethanreal:
Animations are not working Firefox 20.0 because its requestAnimationFrame callback passes the epoch time as an argument while the code expects performance.now(), and the code to wrap the callback to use performance.now() didn't catch this case.

### DIFF
--- a/web-animations.js
+++ b/web-animations.js
@@ -2891,19 +2891,19 @@ if (!raf) {
         callback(clockMillis());
       }, 1000/60);
     };
-  } else if (usePerformanceTiming) {
-    // platform requestAnimationFrame provides only millisecond accuracy, wrap
-    // it and use performance.now()
-    raf = function(callback) {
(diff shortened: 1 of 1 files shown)


morethanreal:
It turns out toolkit normalizes requestAnimationFrame and broke the feature detection in WebAnimations...

However this issue might resurface once mozilla unprefixes requestAnimationFrame.

dstoc:
Won't Firefox only unprefix requestAnimationFrame once it matches the spec?

birtles:
requestAnimationFrame is unprefixed in Firefox Nightly (23) and uses DOMHighResTimeStamp.

Note that the current WD of RAF still refers to DOMTimestamp (the change is only in the ED) so there are likely to be other UAs that provide a DOMTimestamp (incl. ESR versions of Firefox).

Issue #202, opened 2013-05-13
fillMode: none wrapped in fillMode: forwards shouldn't fill forwards


shans:
If an animation finishes right at the end of a container, and the container has a forward fillMode, then the animation fills forward regardless of its own fillMode. This is wrong :)

steveblock:
As I read the spec, this is working as intended. I think the spec makes sense too.

If I have a hierarchy of TimedItems which represents a compound animation, I want to be able to consider that as a black box, and not worry about the details of its internals. So if I want the overall effect of the compound animation to fill forwards, I should be able to do this by setting the fill mode on just the root TimedItem. I don't want to have to walk the tree and update the fill mode of all child animations which align with the end of the root TimedItem.

shans:
At the moment, if the parent container ends exactly when the child does
then the parent full overrides the child. If it ends even a fraction of a
second later then it does not. One of these is clearly a bug :)

With regards to your wider point, the internals define which parts of the
animation should persist and which shouldn't. The container only gets to
decide on whether that persistence carries past the boundary of the parent.

Cheers,
    - Shane

PR #203, opened 2013-05-13
Pass parent filling state to child when updating iteration time.

shans:
This is a fix for #202.

Looking further I wonder why this is so stateful. Can we get away with always passing time values into the sample function and generating animation currentTimes, etc. on request?

### DIFF
--- a/test/testcases.js
+++ b/test/testcases.js
@@ -12,6 +12,7 @@ var tests = [
   'test-duration.html',
   'test-element-animate.html',
   'test-fill-values.html',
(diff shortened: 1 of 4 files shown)


mithro:
We can't merge this automatically, can you merge and then update?

shans:
This should not be merged until the spec is updated

dstoc:
The spec is now updated. Should we go ahead with this patch or refactor to introduce the before/active/after phases?

shans:
Probably refactoring is the better option

Issue #208, opened 2013-05-21
document has no property 'animationTimeline'


wesleycho:
I was going through the demos and I noticed when running `demo_example.html` that in the `web-animations-visualization.js` file, there is usage of `document.animationTimeline`, but this is not referred to anywhere in `web-animations.js` or `web-animations-visualization.js`, or in any history that I could find.

PR #209, opened 2013-05-21
Fixed web-animations-visualization.js's use of document.animationTimeline and use of play() in demo_example.html

wesleycho:
I fixed the mistake in `web-animations-visualization.js` where it uses an outdated method of playing animations.

I also fixed fixed the outdated use of `play()` in `demo_example.html`.

This is meant to fix [#208](https://github.com/web-animations/web-animations-js/issues/208)


steveblock:
Thanks!

PR #211, opened 2013-06-03
Making testcase.js valid JSON.

mithro:
From http://en.wikipedia.org/wiki/JSON
- String (double-quoted Unicode, with backslash escaping)


dstoc:
Why? It's not really JSON, since you wouldn't be able to do "var foo = " there.

mithro:
I can use "simplejson.loads(data[10:-2])" in Python then...

mithro:
I guess we should use the JSONP standard; http://en.wikipedia.org/wiki/JSONP

```
<script type="text/javascript"
        src="http://server2.example.com/Users/1234?jsonp=parseResponse">
</script>
parseResponse({"Name": "Foo", "Id": 1234, "Rank": 7});
```

dstoc:
No need for that.

lgtm if this unblocks your work in Python. Though I'll note this is trivial enough that you could just use a regex and keep this in JS style.

PR #213, opened 2013-06-13
Rewriting the test framework (to be closer to testharness) and porting all tests.

mithro:
My goals where:
- Make the tests more readable;
- Make the tests non-flaky or timing dependent;
- Make the tests faster by removing setTimeout and other slow operations;
- Make basic performance analysis work (by excluding loading time);
- Make our custom test changes possible to send upstream;

The code has been split into three files;
- testharness_styles.js - Contains the extra asserts and code for dealing with styles
- testharness_timing.js - Contains the code to allow testharness deal with time.
- testharness_bootstrap.js - Contains the small remaining code which auto-includes all the testharness pieces

== Adding assert functions for checking style properties ==
First thing I have added is a bunch of assert functions which cover our common
check use case. These are the form of the existing assert_equal and other
assert functions.

```
assert_style(<object>, <style dictionary>);
assert_style(document.getElementById('id'), {'width': '24px'});

assert_dom_style(<index>, <query selector>, <style dictionary>);
assert_dom_style(0, '#id',   {'width': '24px'});
assert_dom_style(5, '.anim', {'width': '24px'});
```

== Automatic test name generation ==
Another small change is that we don't name our tests very often, so I've
extended generate_tests to allow a "null" input name and automatically create
them.

```
generate_tests(assert_equals, [
    [null, 1+1, 2],
    [null, 1+0, 1]
    ])
```

equivalent to;

```
generate_tests(assert_equals, [
    ["assert_equals(2, 2)", 1+1, 2],
    ["assert_equals(2, 2)", 1+0, 1]
    ])
```

It even does a smarter job with assert_style and assert_dom_style giving the
current names like;
    An anonymous div assert_style({'width': '24px'})
or
    objectid assert_style({'width': '24px'})

They could be made even nicer....

== Teaching test-harness.js about timing ==

To replace the setTimeout and .currentTime changing I needed to teaching
test-harness.js about timing. I did this by adding two new functions which
mirror the original test and generate_tests functions. Internally it uses
testharness's async_tests and a custom time tracking system so they all appear
in the testharness report as expected.

test_at function
    test(<function>, <name>);
    test_at(<time to test at>, <function>, <name>);

For example;
    test(function() {assert_true(true)}, "assert_true with true")
    test_at(0.5, function() {assert_true(true)}, "assert_true with true at 0.5")

generation_tests_at function
    generate_tests(<test_function>, <list of parameters>, <properties>, <common parameters>)
    generate_tests_at(<test_function>, <list of parameters>, <properties>, <common parameters>)

For example;
    generate_tests(assert_equals, [
         ["Sum one and one", 1+1, 2],
         ["Sum one and zero", 1+0, 1]
         ])
    generate_tests_at(0.5, assert_equals, [
         ["Sum one and one", 1+1, 2],
         ["Sum one and zero", 1+0, 1]
         ])

The existing code is changed in the following ways;

=== Changes to *-checks.js ===

The old code which reads;
    check(document.querySelectorAll('.anim')[0],{'width':'50px'},0.5);
    check(document.querySelectorAll('.anim')[1],{'width':'50px'},0.5);
    check(document.querySelectorAll('.anim')[2],{'width':'50px'},0.5);

becomes
    test_at(0.5, function(){ assert_dom_style('.anim', 0, {'width':'50px'}); }, null);
    test_at(0.5, function(){ assert_dom_style('.anim', 1, {'width':'50px'}); }, null);
    test_at(0.5, function(){ assert_dom_style('.anim', 2, {'width':'50px'}); }, null);

which becomes
    generate_tests_at(0.0, assert_dom_style, [
        [null, 0, '.anim', {'width':'50px'}],
        [null, 1, '.anim', {'width':'50px'}],
        [null, 2, '.anim', {'width':'50px'}]
        ]);

=== Changes to setTimeout tests ===
For setTimeout tests the changes are less drastic and more readable.

var animFunc = {left: ["100px", "200px"]};

// Test that the animation interval uses inclusive start and end bounds.
var elem1 = document.getElementById("anim1");
var player1 =
    document.timeline.play(new Animation(elem1, animFunc, 1.0));

test_at(
    0.0,
    function() { assert_equals(getComputedStyle(elem1).left, "100px") },
    "Start bound should be inclusive");

test_at(
    1.0,
    function() { assert_equals(getComputedStyle(elem1).left, "200px") },
    "End bound should be inclusive");


shans:
I know it's a pain, but I'd like to see the following changes to this:

(1) generate_tests_at should accept a single query selector as an argument, and use the position of the test data in the list as the index into the provided query selector results, rather than requiring an explicit index.
(2) the test data in test_at should have the test name last, so that the null can simply be elided.
(3) the test data should not need to be a list if no test name is specified

 i.e. instead of:

generate_tests_at(0.0, assert_dom_style, [
[null, 0, '.anim', {'width':'50px'}],
[null, 1, '.anim', {'width':'50px'}],
[null, 2, '.anim', {'width':'50px'}]
]);

I'd like to see:

generate_tests_at(0.0, ".anim", [{'width':'50px'}, {'width':'50px'}, {'width':'50px'}]);

I realize that this is less flexible:
(a) we can't roll multiple query selectors into a generate_tests_at call
(b) we can't pass a custom assert method into generate_tests_at - effectively, assert_dom_style is always used.

I'm OK with these restrictions though, because I think the above is very idiomatic.

A final change:
(4) assert_dom_style should do the same thing - i.e. be assert_styles instead, and take a list of styles that match dom indices.

If we decide to go this way, these changes wouldn't necessarily need to be done before commit - I'd be happy with a follow-up patch that implements them.

Thoughts?

dstoc:
I'm confused why assert_dom_style is still here. My suggestion was not to include it.

shans wrote:

> generate_tests_at(0.0, ".anim", [{'width':'50px'}, {'width':'50px'}, {'width':'50px'}]);

Agree in general, but,
Given the pattern in testharness, this might be better named something like 'generate_style_tests_at'.
The third argument should either be a list, or a single object to apply against all matching elements.
The second argument should be either an element, or a node list. Either pass $('.anim') or $$('.anim') -- it should not run the querySelector logic itself.

assert_style should work the same way

mithro:
So, one of the guiding principles I was using to do this rewrite is;
 **Make the testing infrastructure consistent with the existing testharness code**

Hence, existing generate_tests takes a test name as the first argument in the list, generate_tests_at should also take the test name as the first argument, generate_tests requires a list, so generate_tests_at requires a list, etc

I think shans is also mixing the timing code with the style code. In further refactoring I'd like to make that even more separated by having the raf overloading be a separate class you give to the TestTimeline.

Isn't $('.anim') or $('.anim') , jquery specific?

shans:
As Doug suggested, if we don't call generate_tests_at generate_tests_at, then we can improve the behavior at the same time.

Fundamentally, generate_tests is not very useful. Extending that to generate_tests_at isn't a good thing to do. Looking at our use cases, the automatically generated expectations don't need to be more compressed (why bother?), which leaves manual testing.

I can see a reasonable argument for something that takes a selector, a time, and a list of style requirements. I don't really mind what it's called.

mithro:
Building our own primitives on top of generate_test_at / generate_test is possible, but again I think strong consistency with the current testharness should be a goal. Specially if we want to push our work back upstream. The further we move away from their API the harder it will be to do any of that.

As mentioned I don't think we should be mixing the style / timing functionality either.

The change to generate_test_at was not to make it more compact, but to make it more consistent.

I'm open to changes too assert_style / assert_dom_style - but again they should be consistent with the existing assert functionality and be generic rather than specific to our case.

mithro:
From discussion with shans;

Replace `test_at` with a 

```
timed_test(null, function() {
   at(1.0, function() {});
   at(1.0, assert_xyz, arguments);
}
```

mithro:
assert_style(<node>, <dictionary>)

Replace
`assert_dom_style(<number>, <query selector>, <dictionary>)`
with
`assert_styles(<query selector>, <dictionary> or <list of dictionaries>)`
When single dictionary, check every node has the given style. When list of dictionary, first node has first dictionary style....

mithro:
Check styles becomes closer too;

```
timed_test(name, 
 function() {
   at(1.0, assert_styles, '.anim', [{}, {}, {}]);
```

dstoc:
Can we make assert_styles also accept elements directly? Something like:

```
assert_styles(<element> or <list of element> or <node list> or <query selector>, <dictionary> or <list of dictionaries>)
```

mithro:
Yes that is possible. What is the correct way in javascript to check if its
an element verse list verses string?

mithro:
assert_styles API

``` javascript
     * asserts that elements returned from a query selector have a list of styles.
     *
     * Assert the element with id #hello is 100px wide;
     *     assert_styles(document.getElementById('hello'), {'width': '100px'})
     *     assert_styles('#hello'), {'width': '100px'})
     *
     * Assert all divs are 100px wide;
     *     assert_styles(document.getElementsByTagName('div'), {'width': '100px'})
     *     assert_styles('div', {'width': '100px'})
     *
     * Assert all objects with class "red" are 100px wide;
     *     assert_styles(document.getElementsByClassName('red'), {'width': '100px'})
     *     assert_styles('.red', {'width': '100px'})
     *
     * Assert first div is 100px wide, second div is 200px wide;
     *     assert_styles(document.getElementsByTagName('div'),
     *                   [{'width': '100px'}, {'width': '200px'}])
     *     assert_styles('div',
     *                   [{'width': '100px'}, {'width': '200px'}])
     *
     * @param {string|Element|Array.<Element>} objects Either;
     *      * A query selector to use to get DOM nodes,
     *      * A DOM node.
     *      * A list of DOM nodes.
     * @param {Array.<Object.<string, string>>|Object.<string, string>} style
     *     See _assert_style_get for information.
```

PR #214, opened 2013-06-17
rename getPlayer() to player()

mikelawther:


shans:
player() needs to be a property too - should just be a matter of using the 'get' syntax. I'm not sure whether player is settable or not - you'll need to check the spec.

PR #221, opened 2013-06-18
Remove unused code and fix some stylistic issues in testharness_extetensions/timing

dstoc:


dstoc:
assert_transform? I grepped, there appear to be no callers.

PR #222, opened 2013-06-18
Implement new Timing interface and rename TimedItem.timing to specified 

dstoc:


dstoc:
Still need to update/add some tests, but ready for a look.

shans:
LGTM so far.

TimedItem.animationTime and TimedItem.iterationTime aren't exposed any more? Can we use TimedItem.currentTime or whatever the exposed time is in the tests instead? Or maybe convert/split the tests so that only unit tests look at the internal values?

dstoc:
Removed uses of iterationTime and currentTime from tests.
Fixed failing tests.

PR #223, opened 2013-06-18
zero() functions do not need an input value, just a type

steveblock:


mithro:
Probably not the right place to discuss, but Isn't zero always unitless?

PR #249, opened 2013-06-19
Change the test runner to load tests in parallel and then run them.

mithro:
This change;
- makes our tests run in less than 10 seconds. 
- failing tests are much more apparent and are less likely to sneak through - they take 5.0 seconds to timeout now.

### DIFF
--- a/test/test-runner.html
+++ b/test/test-runner.html
@@ -3,8 +3,6 @@
 
 vim: set ts=2 sw=2 et sts=2 ai:
 
-vim: set expandtab shiftwidth=2 tabstop=2:
-
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
--- a/test/test-runner.html
+++ b/test/test-runner.html
@@ -207,7 +207,7 @@ function testStarted(test) {
  */
 function testFinished(test) {
   test.finished = now();
-  test.querySelector('.timing').innerText = (test.finished - test.start).toFixed(2) + 'ms';
+  test.querySelector('.timing').textContent = (test.finished - test.start).toFixed(2) + 'ms';
 }
 
 /* @type {?number} */ var intervalId = null;
@@ -342,20 +342,20 @@ function changeTestState(test, newState) {
  * Update the status dialog with information about the current status.
  */
 function updateStatus() {
-  loadingElement.innerText = loadingTests.length;
-  unstartedElement.innerText = unstartedTests.length;
-  runningElement.innerText = runningTests.length;
-  postingElement.innerText = postingTests.length;
-  finishedElement.innerText = finishedTests.length;
+  loadingElement.textContent = loadingTests.length;
+  unstartedElement.textContent = unstartedTests.length;
+  runningElement.textContent = runningTests.length;
+  postingElement.textContent = postingTests.length;
+  finishedElement.textContent = finishedTests.length;
 
   if (loadingElement.length > 0) {
-    statusElement.innerText = 'Loading';
+    statusElement.textContent = 'Loading';
   } else if (runningTests.length > 0) {
-    statusElement.innerText = 'Running';
+    statusElement.textContent = 'Running';
   } else if (postingTests.length > 0) {
-    statusElement.innerText = 'Posting results';
+    statusElement.textContent = 'Posting results';
   } else if (finishedTests.length == tests.length) {
-    statusElement.innerText = 'Finished';
+    statusElement.textContent = 'Finished';
     window.clearInterval(statusIntervalId);
   }
 }
@@ -532,7 +532,7 @@ function processResults(test, results) {
     }
 
     var count = test.querySelector('.count');
-    count.innerText = passed + ' of ' + results.length + ' passed';
+    count.textContent = passed + ' of ' + results.length + ' passed';
 
     var log = test.querySelector('.log');
     return testResult;
--- a/test/test-runner.html
+++ b/test/test-runner.html
@@ -1,5 +1,7 @@
 <!--
-Copyright 2012 Google Inc. All Rights Reserved.
+@license Copyright 2012 Google Inc. All Rights Reserved.
+
+vim: set ts=2 sw=2 et sts=2 ai:
 
 vim: set expandtab shiftwidth=2 tabstop=2:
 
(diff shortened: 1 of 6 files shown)


test/test-runner.html
```diff
@@ -1,5 +1,7 @@
 <!--
-Copyright 2012 Google Inc. All Rights Reserved.
+@license Copyright 2012 Google Inc. All Rights Reserved.
+
+vim: set ts=2 sw=2 et sts=2 ai:
```
shans: Probably should remove this so Doug doesn't get mad

test/test-runner.html
```diff
@@ -1,7 +1,7 @@
 <!--
-Copyright 2012 Google Inc. All Rights Reserved.
+@license Copyright 2012 Google Inc. All Rights Reserved.
 
-vim: set expandtab shiftwidth=2 tabstop=2:
+vim: set ts=2 sw=2 et sts=2 ai:
```
shans: Move to bottom

test/test-runner.html
```diff
@@ -1,7 +1,7 @@
 <!--
-Copyright 2012 Google Inc. All Rights Reserved.
+@license Copyright 2012 Google Inc. All Rights Reserved.
```
dstoc: remove @license

test/test-runner.html
dstoc: remove this one, since it's the default

test/test-runner.html
```diff
@@ -17,90 +17,450 @@
...
+  border: 1px solid black;
+  padding-right: 5px;
+}
+
+h1 {
+  display: inline;
+}
+
+table {
+  border-collapse:collapse;
```
dstoc: Space after :

test/test-runner.html
```diff
@@ -17,90 +17,450 @@
...
-var currentTest = 0;
-var currentTestStart;
 
-var testCasesRun = 0;
-var testCasesPassed = 0;
 
-var runID = window.location.href.split("?")[1];
-var checkForResultsTimeout;
+/**
+ * Get the most accurate version of time possible.
```
dstoc, line 104: performance.now() would be the most accurate version of time possible, should we just remove this method and call performance.now()?

test/test-runner.html
```diff
@@ -110,152 +470,78 @@
...
     if (results[x].status == 0) {
       grade = "pass";
     }
-    newResultsDiv.innerHTML += grade + " ";
+    output.className = grade;
+
+    output.innerHTML += grade + " ";
     if (results[x].message != null) {
-      newResultsDiv.innerHTML += results[x].message;
+      output.innerHTML += results[x].message;
```
dstoc, line 504: Concat strings, then append to innerHTML in a single call.

test/test-runner.html
```diff
@@ -110,152 +470,78 @@
...
-    if (childs[x].nodeName == "DIV"){
-      if (childs[x].style.display == 'block') childs[x].style.display = 'none';
-      else {
-        childs[x].style.display = 'block';
-        changeIframe = true;
-      }
+  logElement.appendChild(newResultsDiv);
+
+  if (results.length > 0) {
+    var testResult = {};
```
dstoc: just initialize it inline:
var testResult = {
  type: 'result',
  ...
}

test/test-runner.html
```diff
@@ -110,152 +470,78 @@
...
+    testResult["type"] = "result";
+    testResult["testName"] = test.id;
+    testResult["results"] = results;
+
+    var passed = results.filter(function(result) {
+        return result.status == 0;
```
dstoc: find to indent this as

```
    foo(function() {
      bar...
    });
```

test/test-runner.html
```diff
@@ -110,152 +470,78 @@
...
+
+    if (!failed) {
+      test.className = "pass";
+    } else {
+      test.className = "fail";
+
+      // Open up the window of the failed result
+      test.querySelector('.info-row').click();
+      // Scroll to it.
+      window.scroll(0,trueOffsetTop(test));
```
dstoc: space after ,

test/test-runner.html
```diff
@@ -17,90 +17,450 @@
...
+ * +--------------------------+
+ * | iFrame containing test   |
+ * +--------------------------+
+ *
+ * @param {String} testName Name of the test suite being run.
+ * @param {String} testURL The url of the test suite.
+ * @return {Element} tbody containing newly created table rows.
+ */
+function createTestRows(testName, testURL) {
+  var tablegroup = document.createElement("tbody");
```
dstoc: FYI, we should really be using single quoted strings throughout. I realize there's a lot of existing inconsistency.

test/testharness_timing.js
```diff
@@ -514,14 +518,44 @@
...
+    // FIXME
+    if (/start=disable/.test(window.location.hash)) {
+        window.testharness_timeline = {"schedule": function(f, t) { setTimeout(f, t) }};
+        window.test_at = function() {};
+        return;
+    }
 
     function testharness_timeline_setup()
     {
+
```
dstoc: extra blank line

test/test-runner.html
```diff
@@ -17,90 +17,450 @@
...
+    finishedTests.unshift(test);
+    haveAllTestsFinished();
+    break;
+  }
+
+  function testSort(a, b) {
+    return a.id.localeCompare(b.id);
+  }
+
+  // Sort the lists just to be nice.
```
dstoc: Is being nice the only purpose of sorting?

test/testharness_timing.js
```diff
@@ -1,5 +1,7 @@
 /**
- * Copyright 2013 Google Inc. All Rights Reserved.
+ * @preserve Copyright 2013 Google Inc. All Rights Reserved.
```
dstoc: remove @preserve

test/testharness_timing.js
```diff
@@ -1,5 +1,7 @@
...
+ *
+ * vim: set expandtab shiftwidth=4 tabstop=4:
```
dstoc: move to bottom

dstoc:
lgtm, but see nits above

test/test-runner.html
mithro: Done.

test/test-runner.html
mithro: If we ever move to a compiler we'll need to add it back otherwise it'll strip the license....

test/test-runner.html
mithro: Done.

test/test-runner.html
mithro, line 104: This might cause problems when we move to supporting firefox/ie/etc?

test/test-runner.html
mithro, line 504: It only occurs 3 times here, I don't think it's worth optimising.

test/test-runner.html
mithro: Done.

test/test-runner.html
dstoc, line 104: Sure, then let's update this function do do what the comment says, use performance.now if available, otherwise Date.now.

test/test-runner.html
mithro: You sure that is right? Shouldn't the foo( cause 1 level of indent and then the function another?

test/test-runner.html
mithro: Done.

test/test-runner.html
mithro: Single quotes for single characters and double quotes for strings, like C?

mithro: I ended up going with double quotes throughout...

test/testharness_timing.js
mithro: Done.

test/test-runner.html
mithro: and provide a reliable execution order

test/testharness_timing.js
mithro: See above about @license - I guess should use @license here for consistency...

test/testharness_timing.js
mithro: Done.

test/test-runner.html
dstoc: Yes, see "Passing anonymous functions" in the style guide.

test/test-runner.html
dstoc: Please follow the style guide. Prefer single quotes.

test/test-runner.html
mithro, line 104: Done.

test/test-runner.html
dstoc: http://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml?showone=Strings#Strings

test/test-runner.html
dstoc: That's what the comment should say.

test/test-runner.html
mithro: Fixed.

test/testharness_timing.js
dstoc: No, for consistency remove it everywhere. We're certainly not going to use a compiler for tests.

test/test-runner.html
mithro: Already done...

test/testharness_timing.js
mithro: I have removed it, but the compiler can also be a lint/syntax checker so it good idea to include them generally...

dstoc: Please say "Done." or explain the changes you actually made when you reply. That will shorten the review loop.

test/test-runner.html
mithro: Done.

test/testharness_timing.js
mithro: What is wrong with 

> I have removed it, but the compiler can also be a lint/syntax checker so it good idea to include them generally...

test/test-runner.html
```diff
@@ -17,90 +15,450 @@
...
+}
+
+/**
+ * Called when a test is finished.
+ * @param {object} test dom object representing the test. id will contain the
+ *    test name.
+ */
+function testFinished(test) {
+  test.finished = now();
+  test.querySelector('.timing').innerText = (test.finished - test.start).toFixed(2) + 'ms';
```
dstoc: Firefox does not implement innerText, use textContent instead (throughout).

test/test-runner.html
```diff
@@ -17,90 +15,450 @@
...
+}
+
+/* @type {?number} */ var intervalId = null;
+
+/**
+ * Checks all the tests are in the unstarted state and then kicks of running
+ * the tests.
+ */
+function haveAllTestsLoaded() {
```
dstoc: Should this be called runTestsIfLoaded? "have" sounds like a boolean getter

test/test-runner.html
mithro: Done.

test/testharness_timing.js
dstoc: I was refering to when you said:

> See above about @license - I guess should use @license here for consistency...

That sounded like you were changing this to be @license. You could have just said:

> Done.

test/test-runner.html
mithro: Done.

test/testharness_timing.js
mithro: Done...

PR #253, opened 2013-06-19
Rename manual tests to use manual-test- prefix and update update-testcases script

alancutter:


alancutter:
These tests are passing according to our test harness though I can see visually they're failing.
I've added a blacklist for tests like these, I think it's much more clear than having the tests not appear in a list.

alancutter:
The comments in this pull request were eaten after I rebased.

Running tests that are expected to fail is a good idea though that would be for another patch.
For this patch I'll just rename failing-testcases.jsonp to disabled-testcases.jsonp and keep the same behaviour.
Added comments explaining why tests are disabled.

test/update-testcases.py
```diff
@@ -0,0 +1,28 @@
...
+
+def list_dir(dir_path, filter_func):
+    return sorted(filter(filter_func, os.listdir(dir_path)), key=get_name)
+
+def main():
+    test_dir = os.path.dirname(os.path.realpath(__file__))
+    testcase_dir = os.path.join(test_dir, 'testcases')
+    with open(os.path.join(test_dir, 'testcases.jsonp'), 'w') as testcases_file:
+        testcases_file.write('// This file is automatically generated by test/update-testcases.py.\n')
+        testcases_file.write('// Disable tests by adding them to test/failing-testcases.jsonp.\n')
```
dstoc: disabled-testcases

test/test-runner.html
```diff
@@ -76,8 +76,18 @@
 'use strict';
 var tests = [];
 function testCases(new_tests) { tests = new_tests; };
+function disabledTestCases(disabled_tests) {
+  // Remove failing tests from tests array, Javascript syle.
```
dstoc: You don't really need this comment.

dstoc:
lgtm, but see above

PR #259, opened 2013-06-20
Fix update tests, pep8 cleanup and making travis fail on out of date testcases.jsonp

mithro:


test/update-testcases.py
```diff
@@ -1,28 +1,60 @@
...
+    def is_testcase_file(filename):
+        return (
+            fnmatch(filename, "*.html") and
+            not fnmatch(filename, "manual-test*") and
+            filename not in disabled_tests)
+
+    new_testcases = StringIO.StringIO()
+    new_testcases.write("""\
+// This file is automatically generated by test/update-testcases.py.
+// Disable tests by adding them to test/disabled-testcases.jsonp.
```
alancutter: This should just be test/disabled-testcases without the extension.

.travis.yml
```diff
@@ -11,6 +11,7 @@ install:
   - ./.travis-setup.sh
 
 script:
+  - python ./test/update-testcases.py || (echo "Please run update-testcases!"; false)
```
alancutter, line 14: Can this step be moved into run-tests.py?

test/update-testcases.py
mithro: Fixed.

.travis.yml
mithro, line 14: I don't think it makes sense?

alancutter, line 14: Then we can remove test/testcases.jsonp from the repository and not have to worry about running it manually. Is it used outside of run-tests.py?

mithro, line 14: Yes - its used by the test runner as it has no filesystem access....

alancutter, line 14: I mean have run-tests.py generate test/testcases.jsonp everytime it is invoked by calling on test/update-testcases.py. This way we don't need to run it manually when we add new tests.
We can .gitignore test/testcases.jsonp to keep the file from getting back into the repo.

mithro, line 14: Patches welcome? :)

alancutter, line 14: Should this be using the --update option?

mithro, line 14: No - it's not suppose to do the update, just report the diff and and fail
if there is a diff.

Tim

PR #261, opened 2013-06-20
Revert testscase list back to js

dstoc:


shans:
Merging on the basis that where this is any different, it's shorter and easier to understand.

This also opens up the opportunity for programmatic population of tests in the future. For example, in SVG, many of the tests are numbered (some-test-area-1, some-test-area-2, etc.).

PR #266, opened 2013-06-20
Don't add the reference element to an absent parent

dstoc:
If added to a parent, ensure it is still actually removed when an error
occurs.


shans:
Add to list of test cases too?

dstoc:
Added to testcases.js

PR #267, opened 2013-06-20
Repeat ticks when animation state changes

dstoc:


dstoc:
Came up with something that works, but we need to discuss.

dstoc:
This should work for everything I can think of, except KeyframeAnimationEffect which we can update after the new API has landed.

test/testcases/test-update-state.html
```diff
@@ -0,0 +1,92 @@
...
+  });
+}, 'Updated style should be visible after change to animation tree');
+
+timing_test(function() {
+  at(3, function() {
+    var anim = element.animate({color: ['red', 'green', 'red']}, 1);
+    // TODO: Really this should already be in effect, since the start
+    //       time == lastTick. But if that is fixed it's unclear how
+    //       to start an animation after the last tick (in this
+    //       harness).
```
shans, line 68: What are the use cases for starting an animation after the last tick?

web-animations.js
```diff
@@ -3005,14 +3081,38 @@ var cachedClockTime = function() {
   return cachedClockTimeMillis / 1000;
 };
 
-var ticker = function(rafTime) {
+
+// These functions should be called in every stack that could possibly cause a
+// subsequent tick with the last tick time to apply different effect results.
```
shans: maybe "These functions should be called in every stack that could possibly modify the
effect results that have already been calculated for the current tick"?

web-animations.js
```diff
@@ -3005,14 +3081,38 @@ var cachedClockTime = function() {
...
+var modifyCurrentAnimationStateDepth = 0;
+var enterModifyCurrentAnimationState = function() {
+  modifyCurrentAnimationStateDepth++;
+};
+var exitModifyCurrentAnimationState = function(shouldRepeat) {
+  modifyCurrentAnimationStateDepth--;
+  if (modifyCurrentAnimationStateDepth == 0 && shouldRepeat) {
```
shans, line 3099: Only the outer shouldRepeat counts at the moment. Should all of them count?

dstoc, line 3099: Yes, it's an optimization to only call once when we exit the stack. I believe (excluding animation effects) we can only be in the context of a single player, so if anywhere is true, the outer should be true.

I could track whether shouldRepeat was set elsewhere and assert that it's set for the outer-most too, should I?

test/testcases/test-update-state.html
dstoc, line 68: Not sure if we need to allow it to take effect. But can script even be in a state where the timeline's current time is the same as at the last sample? Currently in the test harness we are always in such a state :(

dstoc, line 68: And no, I don't think there's a good use case for this. It's just weird.

web-animations.js
shans, line 3099: I think that would make me happiest. Adding your justification as a comment works for me too.

test/testcases/test-update-state.html
shans, line 68: Two ways, both kinda equivalent in effect.
1) wrap rAF
2) use a custom effect

Maybe we should be wrapping rAF to test the custom effect mechanism, then implementing timingTests as custom effects.

dstoc, line 68: Not following entirely. My question was about a native implementation, if we can't get into that state then we don't need to decide what should happen.

I thought we might fix the harness by having it simulate a small advance in time after sampling, before running the checks.

web-animations.js
dstoc: Done.

web-animations.js
dstoc, line 3099: Done (comment and assertion).

dstoc, line 3099: Not quite done, since it fails :)

dstoc, line 3099: Done (the assertion actually works now).

dstoc, line 3099: I was wrong, we can't assert that since I violate it on purpose ;) The constructor for PathAnimationEffect calls exit with shouldRepeat = false, because it internally calls the angle/autoRotate setters which set shouldRepeat = true. The setters can't tell which player the effect is attached to so it's set unconditionally.

So we can keep a comment about that, but not an assertion.

PR #269, opened 2013-06-21
Remove bit-rotted web-animations-test

dstoc:


shans:
Ah these were the good(?) old days. Templates, groups in arguments, 'Anim's. It's a right walk down memory lane I tell you.

PR #270, opened 2013-06-21
Mainly automatic transform of test_at to the timing_test(function() { at() }) API.

mithro:
Because shans said so....


mithro:
No comments I guess?

PR #271, opened 2013-06-21
Adding ability to disable testharness without breaking the page.

mithro:


mithro:
@shans @steveblock - How does this look?

test/testharness_stub.js
```diff
@@ -0,0 +1,72 @@
...
+            if (!(components[i] in target))
+            {
+                target[components[i]] = {};
+            }
+            target = target[components[i]];
+        }
+        target[components[components.length - 1]] = object;
+    }
+    /* ********************************************************************* */
+    // Expose all the test functionality as empty stubs
```
shans, line 40: If we resolve on removing this we should remove it from here too.

shans:
How do you switch between stub and bootstrap?

mithro:
You put a testharness=disable into the URL somewhere - probably in either the query string or hash string....

/testharness=disable/.test(window.location.href)

test/testharness_stub.js
mithro, line 40: Doug would like it removed, I'd like to keep it - yet to have a discussion about which way we should go.

shans:
I think we discussed before the possibility of having a button to switch into manual mode, like we used to? (Actually I think it used to be a dropdown list option).

It doesn't seem like this approach makes that possible without reloading the page, as you actually need to switch out one library for another, right?

mithro:
You are correct, this doesn't support stubbing out testharness without a reload. It does however allows you to load a test without having the tests interfere with the page.

The other change I think what you are asking for is a switch between "run time at real time" and "run time as fast as possible skipping time".

shans:
Yes, you think correct.

That change is actually more useful than this one, I think? For example today I was getting strange behaviour in a test and wanted to just quickly see what was really going on.

Could you get the test harness to add a link 'run with testharness disabled' (and 'run with testharness enabled' if the stub is loaded) as a substitute? This would probably be as convenient.

mithro:
Yes, adding that link would be possible. Will do now.

Could you explain the below better;

> strange behaviour in a test and wanted to just quickly see what was
> really going on.

What is strange behaviour? What made you think testharness was involved?

shans:
The specific strange behaviour I was looking at isn't relevant; however here it is: my animations seemed to be iterating half as many times as requested going by the event logs.

I didn't think testharness was involved.

mithro:
Specific strange behaviour gives me a better idea behind the "just run the
animation as normal" use case. I'm sure there are real reasons that it is a
good idea but I'm having trouble seeing them?

One a side note, when run under run-tests we could do "reload on file
change" type behaviour which would allow you to run in split screen, make a
change and then see the animation instantly reload....

shans:
The most general way I can think to put it is that when implementing new features you often don't know whether your test cases are broken or the feature code is. Running the feature tests in automatic mode is not helpful under these circumstances.

Suffice it to say that both Steve (while implementing the new keyframes feature) and I (while implementing the new events feature) have repeatedly expressed a desire to be able to do this :)

mithro:
> Suffice it to say that both Steve (while implementing the new keyframes
> feature) and I (while implementing the new events feature) have repeatedly
> expressed a desire to be able to do this :)

Both you and Steve have expressed a desire to "run tests in manual mode".
I've only got a cursory memory of the previous "manual mode" and still
don't quite understand which part of the feature is the actual important
part.

The first time you mentioned "manual mode" I thought you meant being able
to change the time to arbitrary positions and stop there (which we can
already do). Then I thought it was disabling tests code, which again was
not quite what you wanted. Hence I'm obviously still trying to figure out
what that actually means.

Your "manual mode" seems to be connected with the idea of an animation
which takes 1 second actually taking 1 real world second? I would describe
that as a "real time mode". I can trivially implement this mode but the
name tells me nothing about how the rest of the system should function, IE
if the tests should run, what should happen when "time runs out", etc.

> The most general way I can think to put it is that when implementing new
> features you often don't know whether your test cases are broken or the
> feature code is.

If the above was /actually/ your goal, then having the ability remove all
test code seems like an answer to me? Then the only place it could be
broken is your code as the test code doesn't exist.

Other things which I could see helping with the above problems include;
- Being able to stop an arbitrary t=X (could be done with creating an
  empty test right now and then clicking on the correct bubble)
- Being able to break into the debugger at a given t=X
- Being able to restart the animation and events from t=0 (without
  reloading the webpage)

test/testharness_stub.js
```diff
@@ -0,0 +1,72 @@
...
+
+    // The assert_xxx functions are not needed as they should only be used
+    // inside test() functions which we have already stubbed out.
+
+    // testharness_extensions.js
+    // Currently only defines assert_XXX functions.
+
+    // testharness_timing.js
+    expose(new_test_stub, 'timing_test');
+    expose(function(seconds, f) { setTimeout(f, seconds*1000.0); }, 'at');
```
mithro, line 97: I don't think this at() function is ever called because the body of the timing_test method is never called...

Actually, it seems like it would be a bad idea to mutate state inside the at() function inside a timing_test... but I think we do actually do that...

mithro:
This pull request depends on https://github.com/web-animations/web-animations-js/pull/305 to provide the GUI hooks.

PR #274, opened 2013-06-22
Resolve some outstanding issues in test-runner and bootstrap.

dstoc:
- Rename bootstrap. As discussed it's unlikely we would upstream this file.
- Load the polyfill automatically (in preparation for client side coverage instrumentation).
- Add and use flags for nochecks and nopolyfill.
- Remove console logging.
- Stop trying to post results when the server doesn't support it.

### DIFF
--- a/test/bootstrap.js
+++ b/test/bootstrap.js
@@ -13,21 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-"use strict";
+'use strict';
 
-// Load the dependencies in order
-var loadScript = function(src)
-    {
-        document.write(
-            '<script type="text/javascript" src="'+ src + '"></script>');
-    };
+(function() {
 
-// Load the required CSS
-var loadCSS = function(src)
-    {
-        document.write(
-            '<link rel="stylesheet" type="text/css" href="' + src + '">');
-    };
+function loadScript(src) {
+  document.write('<script type="text/javascript" src="'+ src + '"></script>');
+}
+
+function loadCSS(src) {
+  document.write('<link rel="stylesheet" type="text/css" href="' + src + '">');
+}
 
 loadScript('../testharness/testharness.js');
 loadCSS('../testharness/testharness.css');
@@ -43,4 +39,5 @@ document.write('<div id="log"></div>');
 loadScript('../testharness/testharnessreport.js');
 
 window.__coverage__ = parent.window.__coverage__;
-// vim: set expandtab shiftwidth=4 tabstop=4:
+
+})();
--- a/test/bootstrap.js
+++ b/test/bootstrap.js
@@ -38,6 +38,8 @@ loadScript(location.pathname.replace('.html', '-checks.js'));
 document.write('<div id="log"></div>');
 loadScript('../testharness/testharnessreport.js');
 
+loadScript('../../web-animations.js');
+
 window.__coverage__ = parent.window.__coverage__;
 
 })();
--- a/test/bootstrap.js
+++ b/test/bootstrap.js
@@ -17,6 +17,8 @@
 
 (function() {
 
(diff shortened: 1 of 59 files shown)


shans:
This looks good to me. I'd like Tim to have a quick look too.

test/bootstrap.js
```diff
@@ -13,21 +13,27 @@
...
-var loadCSS = function(src)
-    {
-        document.write(
-            '<link rel="stylesheet" type="text/css" href="' + src + '">');
-    };
+'use strict';
+
+(function() {
+
+var thisScript = document.querySelector("script[src$='bootstrap.js']");
```
mithro, line 20: What's the style guide for this? Should quotes be double or single?

test/bootstrap.js
```diff
@@ -37,10 +43,17 @@ loadScript('../testharness_extensions.js');
 loadScript('../testharness_timing.js');
 loadCSS('../testharness_timing.css');
 
-loadScript(location.pathname.replace('.html', '-checks.js'));
+if (!isUnitTest() && !hasFlag('nochecks')) {
+  loadScript(location.pathname.replace('.html', '-checks.js'));
+}
```
mithro, line 48: Shouldn't this be or? Unittests shouldn't have -check files right?

test/testharness_timing.js
```diff
@@ -148,7 +148,6 @@
         while (callbacks.length > 0)
```
mithro, line 148: Both these console.log's have been really useful when debugging tests.....

test/bootstrap.js
dstoc, line 48: Condition seems right to me. Which case fails?

test/bootstrap.js
dstoc, line 20: Prefer single quotes. But since this string contains single quotes it's appropriate to use a double quote here.

test/testharness_timing.js
dstoc, line 148: These logs aren't descriptive, it's not clear what's being logged here. They obscure the output from individual test cases.

test/bootstrap.js
mithro, line 48: Yours works thanks to De Morgans, but I find it hard to read what you are actually trying to say which is;

> If a unittest or has nochecks skip loading of -checks.js

Maybe the following is clearer?

If (!(isUnitTest() || hasFlags('nochecks'))) {}

mithro:
Just the two remaining comments, then LGTM.

test/testharness_timing.js
mithro, line 148: So the proper fix it to make the log messages clearer?

test/bootstrap.js
dstoc, line 48: Can you read that back to me? 

I read my version as:

If it is not a unit test and it does not have the "nochecks" flag, then load the checks file.

mithro, line 48: I agree yours reads as that. 

I don't agree that it matches what you're actually trying to say; which I claim is;
- Unit tests don't have -checks files.
- Tests are flagged as nochecks should not load check files.

For unittests, it would be much safer to try too load -checks.js make sure it _fails_. It is really easy to create a checks file for a unittest and then wonder why it is not working (or even worse just assume it is and then have other people copy you!).

test/testharness_timing.js
dstoc, line 148: I don't know what they are trying to log, and they should't be showing up by default.

You can add them back if you like, but right now I'd like to solve the problem that it's very hard to read the console output.

test/bootstrap.js
dstoc, line 48: I'm not sure how changing the condition would fix that. I don't think it's worth trying to detect this, at the cost of more output in the console. We can catch such problems at review time or in update-testcase.py or some other presubmit tool, etc. But any of those should be done in a subsequent change.

PR #287, opened 2013-06-24
Events

shans:


test/testcases/impl-test-player-sees-handlers.html
```diff
@@ -0,0 +1,87 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<!DOCTYPE html>
+<body></body>
+<script src="../testharness_bootstrap.js"></script>
+<script src="../../web-animations.js"></script>
```
dstoc: No longer needed.

test/testcases/impl-test-player-sees-handlers.html
```diff
@@ -0,0 +1,87 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<!DOCTYPE html>
+<body></body>
+<script src="../testharness_bootstrap.js"></script>
```
dstoc: This is now just "../bootstrap.js"

test/testcases/test-events.html
```diff
@@ -0,0 +1,200 @@
...
+div.anim {
+  position: relative;
+  left: 0px;
+  width: 100px;
+  height: 100px;
+  background-color: green;
+}
+</style>
+<div class='anim'></div>
+<script src="../testharness_bootstrap.js"></script>
```
dstoc: As above

web-animations.js
```diff
@@ -655,8 +673,172 @@ TimedItem.prototype = {
...
+      // Did we pass the start of this animation in the reverse direction?
+      } else if (fromTime >= startTime && toTime < startTime) {
+        this.onend(new TimingEvent(constructorToken, this, 
+            this.specified.startDelay, toGlobal(startTime), firstIteration));
+      }
+    }
+  },
 };
 
+var TimingEvent = function(token, target, localTime, timelineTime, iterationIndex, seeked) {
```
dstoc: Need to export this at the bottom of the file.

web-animations.js
```diff
@@ -952,6 +1134,45 @@ TimingGroup.prototype = createObject(TimedItem.prototype, {
         this.localTime + ') ' + ' [' +
         this.children.map(function(a) { return a.toString(); }) + ']'
   },
+  _hasHandler: function() {
+    return TimedItem.prototype._hasHandler.bind(this)() ||
+      (this.children.length > 0 &&
+        this.children.map(function(a) { return a._hasHandler(); }).reduce(
```
dstoc: do we need the map? can we just reduce return a._hasHandler() || ... ?

web-animations.js
```diff
@@ -224,6 +224,8 @@ var Player = function(token, source, timeline) {
   this._hasTicked = false;
 
   this.source = source;
+  this._checkForHandlers();
+  this._lastCurrentTime = this._startTime;
```
dstoc: _eventsLastGeneratedTime ? not sure what 'current' is meant to mean

dstoc: I think _startTime can change before _generateEvents is called, is that OK?

web-animations.js
```diff
@@ -248,6 +250,7 @@ Player.prototype = {
     } finally {
       exitModifyCurrentAnimationState(this._hasTicked);
     }
+    this._checkForHandlers();
```
dstoc: This should be inside the try/finally

web-animations.js
```diff
@@ -952,6 +1134,45 @@ TimingGroup.prototype = createObject(TimedItem.prototype, {
         this.localTime + ') ' + ' [' +
         this.children.map(function(a) { return a.toString(); }) + ']'
   },
+  _hasHandler: function() {
+    return TimedItem.prototype._hasHandler.bind(this)() ||
```
dstoc: TimedItem.prototype._hasHandler.call(this)

web-animations.js
```diff
@@ -952,6 +1134,45 @@ TimingGroup.prototype = createObject(TimedItem.prototype, {
...
+      (this.children.length > 0 &&
+        this.children.map(function(a) { return a._hasHandler(); }).reduce(
+          function(a, b) { return a || b }));
+  },
+  _generateChildEventsForRange: function(localStart, localEnd, rangeStart,
+    rangeEnd, iteration, globalTime, deltaScale) {
```
dstoc: more indentation

web-animations.js
```diff
@@ -952,6 +1134,45 @@ TimingGroup.prototype = createObject(TimedItem.prototype, {
...
+    // If our range is not going in the same direction as the delta then
+    // ignore.
```
dstoc: This comment would be better if it said why it's safe to ignore this case

web-animations.js
```diff
@@ -3146,6 +3367,12 @@ var ticker = function(rafTime, isRepeat) {
     }
   }
 
+  var eventList = [];
```
dstoc: This is not used.

web-animations.js
```diff
@@ -356,6 +359,21 @@ Player.prototype = {
       this.source._getAnimationsTargetingElement(element, animations);
     }
   },
+  _handlerAdded: function() {
+    this._needsHandlerPass = true;
+  },
+  _checkForHandlers: function() {
+    this._needsHandlerPass = this.source !== null && this.source._hasHandler();
+  },
+  _generateEvents: function(eventList) {
```
dstoc: param is not used

web-animations.js
```diff
@@ -655,8 +673,172 @@ TimedItem.prototype = {
     return this.parent === null ? effectivePlaybackRate :
         effectivePlaybackRate * this.parent._netEffectivePlaybackRate();
   },
+  set onstart(fun) {
```
dstoc, line 678: addEventListener is TODO?

dstoc, line 678: I guess we'll also want to inherit from EventTarget if possible.

web-animations.js
```diff
@@ -356,6 +359,21 @@ Player.prototype = {
       this.source._getAnimationsTargetingElement(element, animations);
     }
   },
+  _handlerAdded: function() {
+    this._needsHandlerPass = true;
+  },
+  _checkForHandlers: function() {
```
dstoc, line 365: I think we need to call this on tree mutations. Seems like currently we wont detect a handler if it's already attached to a timed item before it's added to a group.

web-animations.js
```diff
@@ -655,8 +673,172 @@ TimedItem.prototype = {
...
+    return isDefinedAndNotNull(this._startHandler) ||
+      isDefinedAndNotNull(this._iterationHandler) ||
+      isDefinedAndNotNull(this._endHandler) ||
+      isDefinedAndNotNull(this._cancelHandler);
+  },
+  _generateChildEventsForRange: function() { },
+  _toSubRanges: function(fromTime, toTime, iterationTimes) {
+    if (fromTime > toTime) {
+      var revRanges = this._toSubRanges(toTime, fromTime, iterationTimes);
+      revRanges.ranges = revRanges.ranges.map(function(a) { return [a[1], a[0]]; }).reverse();
```
dstoc: This is a little cheaper:
ranges = range.forEach(fun(a) { a.reverse(); };
ranges.reverse();

web-animations.js
```diff
@@ -655,8 +673,172 @@ TimedItem.prototype = {
...
+      revRanges.start = iterationTimes.length - revRanges.start - 1;
+      revRanges.delta = -1;
+      return revRanges;
+    }
+    var currentStart = fromTime;
+    var ranges = [];
+    var skipped = 0;
+    while (iterationTimes[skipped] < fromTime) {
```
dstoc, line 736: Maybe a binary search? Or TODO

web-animations.js
```diff
@@ -655,8 +673,172 @@ TimedItem.prototype = {
...
+  _generateChildEventsForRange: function() { },
+  _toSubRanges: function(fromTime, toTime, iterationTimes) {
+    if (fromTime > toTime) {
+      var revRanges = this._toSubRanges(toTime, fromTime, iterationTimes);
+      revRanges.ranges = revRanges.ranges.map(function(a) { return [a[1], a[0]]; }).reverse();
+      revRanges.start = iterationTimes.length - revRanges.start - 1;
+      revRanges.delta = -1;
+      return revRanges;
+    }
+    var currentStart = fromTime;
```
dstoc: Move these  (ranges, currentStart) before the for loop below where they get used.

web-animations.js
```diff
@@ -655,8 +673,172 @@ TimedItem.prototype = {
...
+      }
+    }
+    ranges.push([currentStart, toTime]);
+    return {start: skipped, delta: 1, ranges: ranges};
+  },
+  _generateEvents: function(fromTime, toTime, globalTime, deltaScale) {
+    function toGlobal(time) {
+      return (globalTime - (toTime - (time / deltaScale)));
+    }
+    toGlobal = toGlobal.bind(this);
```
dstoc: toGlobal doesn't use this?

web-animations.js
```diff
@@ -655,8 +673,172 @@ TimedItem.prototype = {
...
+    var localScale = this.specified.playbackRate;
+    var firstIteration = Math.floor(this.specified.iterationStart);
+    var lastIteration = Math.floor(this.specified.iterationStart +
```
dstoc, line 759: Should this be clipped by activeDuration?

test/testcases/impl-test-player-sees-handlers.html
shans: removed

test/testcases/impl-test-player-sees-handlers.html
shans: changed

web-animations.js
shans: Done

web-animations.js
shans: Yeah we can do this in a single step.

web-animations.js
shans: _lastCurrentTime is the previous _currentTime value. _previousCurrentTime?

shans: I think it's OK for _startTime to change after this line. I need to initialize _lastCurrentTime to something sensible for the first tick.

Maybe we should talk about this a bit though.

web-animations.js
shans: Done.

Seeing this here, I'm a little concerned that try { } finally without catch reporting is going to make modifying the code harder as exceptions are silently swallowed. Do you think we could add some catches?

web-animations.js
shans: Done.

web-animations.js
shans: Actually don't need this check any more as I've improved the way ranges are generated. Any cases that would trigger this are filtered by the range clipping now.

web-animations.js
shans: Removed.

I'd initially intended to sort events before delivering, and a strict interpretation of the spec indicates we must do this. However, I think there are very few cases where it actually matters.

web-animations.js
shans: Removed.

web-animations.js
shans, line 678: addEventListener isn't mentioned in the Web Animations specification. I guess we need to work out whether it should be.

What's EventTarget? I'm not sure how we'd inherit from it in the polyfill.

web-animations.js
shans, line 365: Yep. Added some tests to characterize this - it has to be tree mutation with an animation that has an already-registered handler, in a case where the player has already been attached to the group.

I've added a check in TimingGroup._childrenStateModified to fix this.

web-animations.js
shans: forEach doesn't return anything :)

This is much clearer. Is it also cheaper? Is that because reverse() costs less than the manual reverse?

web-animations.js
shans, line 736: Should actually be able to calculate this. I'll TODO for now.

web-animations.js
shans: Done.

web-animations.js
shans: Ah. Good point.

web-animations.js
shans, line 759: It could be, but I don't think the behaviour will change as clipping by start and end time occurs on the ranges.

web-animations.js
birtles, line 678: addEventListener is defined on EventTarget: http://dom.spec.whatwg.org/#interface-eventtarget

shans, line 678: Ah, got it. And we've defined TimedItem to inherit from EventTarget.

shans, line 678: At least in Chrome, EventTarget doesn't surface in JS. Instead, it is rolled into Node.

I can easily do the same thing (add EventTarget methods manually to TimedItem). I think this belongs in a future patch.

web-animations.js
```diff
@@ -652,8 +670,174 @@ TimedItem.prototype = {
...
+      clippedToTime = Math.max(toTime, startTime);
+    }
+    var subranges = this._toSubRanges(clippedFromTime, clippedToTime,
+      iterationTimes);
+    for (var i = 0; i < subranges.ranges.length; i++) {
+      var currentIter = subranges.start + i * subranges.delta;
+      if (isDefinedAndNotNull(this.oniteration) && i > 0) {
+        var iterTime = subranges.ranges[i][0];
+        this.oniteration(new TimingEvent(constructorToken, this,
+          iterTime - this.startTime, toGlobal(iterTime), currentIter));
```
dstoc: more indentation

web-animations.js
```diff
@@ -652,8 +670,174 @@ TimedItem.prototype = {
...
+      }
+      if (subranges.delta > 0) {
+        var iterFraction = this.specified.iterationStart % 1;
+      } else {
+        var iterFraction = 1 - 
+          (this.specified.iterationStart + this.specified.iterationCount) % 1;
```
dstoc: more indentation

web-animations.js
```diff
@@ -652,8 +670,174 @@ TimedItem.prototype = {
...
+      }
+      this._generateChildEventsForRange(
+        subranges.ranges[i][0], subranges.ranges[i][1],
```
dstoc: more indentation

web-animations.js
```diff
@@ -652,8 +670,174 @@ TimedItem.prototype = {
...
+      iterationTimes);
+    for (var i = 0; i < subranges.ranges.length; i++) {
+      var currentIter = subranges.start + i * subranges.delta;
+      if (isDefinedAndNotNull(this.oniteration) && i > 0) {
+        var iterTime = subranges.ranges[i][0];
+        this.oniteration(new TimingEvent(constructorToken, this,
+          iterTime - this.startTime, toGlobal(iterTime), currentIter));
+      }
+      if (subranges.delta > 0) {
+        var iterFraction = this.specified.iterationStart % 1;
```
dstoc: pull the definition outside the if/else. possibly use ternary to initialize

web-animations.js
dstoc: Can we inherit from Event?

shans: Not really. I think I've got it as close as we can reasonably do.

web-animations.js
shans: done.

web-animations.js
shans: done.

web-animations.js
shans: done.

web-animations.js
shans: done.

web-animations.js
shans: Doug and I decided it would be better to initialize _lastCurrentTime as late as possible. This means that if a player is created and the start time is immediately updated then events won't be generated for the initial seek, which we think is the correct behaviour.

PR #293, opened 2013-06-24
Basic "real time" mode (previous called "manual" mode)

mithro:
Will be done in following CLs;
- restarting time
- "halting" by clicking on timeline
- Adding a "pause" button.
- Adding breakpoint ability.

### DIFF
--- a/test/testharness_timing.js
+++ b/test/testharness_timing.js
@@ -536,6 +536,32 @@
         this.toNextEvent();
     };
 
+    // Capture the real requestAnimationFrame so we can run in "real time" mode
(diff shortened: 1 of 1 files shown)


steveblock:
lgtm

Do we have a policy for when to use separate commits for changes in response to review comments? Personally, I think that if the review introduces a logic change (as is the case here, as opposed to simple refactoring), the commits should be squashed together before merging, so that each merged commit is 'correct' when taken on its own.

mithro:
I don't know of any policy and there are good arguments either way....

mithro:
Squashed as requested.

PR #254, opened 2013-06-20
Update KeyframeAnimationEffect to reflect FPWD

steveblock:
- Update KeyframeAnimationEffect API
- Remove Keyframe from API as this is now just a dictionary
- Update algorithm for normalizing the list of keyframes
- Update algorithm for sampling a KeyframeAnimationEffect
- Handle interpolating between keyframes with different composite operations
- Remove redundant KeyframeList and GroupedAnimationEffect types
- Remove merge composite operation
- Update tests accordingly

PR #295, opened 2013-06-25
Follow-up fixes to updated KeyframeAnimationEffect

steveblock:
See https://github.com/web-animations/web-animations-js/pull/254


dstoc:
Don't merge yet, we need to test whether the zero() is really not needed for some invalid value cases.

steveblock:
> Don't merge yet, we need to test whether the zero() is really not needed for some invalid value cases.
> Have removed the commit that removed zero() methods and will investigate this separately.

PR #297, opened 2013-06-25
Fixing update-testcases and adding the thus missing test.

mithro:


test/update-testcases.py
```diff
@@ -54,7 +54,8 @@ def is_testcase_file(filename):
     for line in lines:
         sys.stdout.write(line)
 
-    sys.exit(not lines)
+    sys.exit(int(len(list(lines)) > 0))
```
dstoc, line 57: Can len return something which is not an integer?

dstoc:
lgtm

test/update-testcases.py
mithro, line 57: It can throw an exception is lines is not something which a **len** method...

mithro:
Can you merge?

test/update-testcases.py
dstoc, line 57: No need to call int() then?

mithro, line 57: sys.exit needs an int, len(list(lines)) > 0 returns a boolean..

PR #299, opened 2013-06-25
Null target animations

rjwright:
Allow animations to have null targets.

### DIFF
--- a/test/testcases/test-null-target.html
+++ b/test/testcases/test-null-target.html
@@ -1,0 +1,57 @@
+<!--
+Copyright 2012 Google Inc. All Rights Reserved.
+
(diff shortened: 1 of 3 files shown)


test/testcases/test-null-target.html
```diff
@@ -0,0 +1,57 @@
...
+    left: 0px;
+    border-radius: 50%;
+    background-color: lightsteelblue;
+    }
+</style>
+<div class="anim" id="a"></div>
+<script src="../testharness_bootstrap.js"></script>
+<script src="../../web-animations.js"></script>
+<script>
+  "use strict";
```
dstoc, line 34: For future reference, no need to indent the script, and prefer single quotes to double quotes.

PR #302, opened 2013-06-26
Remove media reference

alancutter:
MediaReference is not part of FPWD. This functionality will be introduced as a separate spec.


dstoc:
Merged manually.

PR #305, opened 2013-06-26
Adding a setting framework to the test-harness.

mithro:
settings

### DIFF
--- a/test/bootstrap-post.js
+++ b/test/bootstrap-post.js
@@ -1,0 +1,32 @@
+/**
+ * Copyright 2013 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


mithro:
JSON.parse("sdfsadf"); doesn't return "sdfsadf" like this function does. I guess I could do a 

``` javascript
var output = null;
try {
  output = JSON.parse(input);
} catch {
  output = input;
}
```

What do you think?

shans:
This change adds significant complexity, but I don't see why we need a settings
framework at the moment.  Something Doug, Steve and I have fallen into the
habit of doing is whiteboarding more complicated changes before starting to
code them. Perhaps we should do that for settings?

mithro:
I didn't really understand this problem until I had a possible solution, it came out of trying to add a GUI to control the various features we want to support. I'm happy to discuss this change further on the whiteboard if you want.

mithro:
RE: Why I ended up creating this;
- We currently have 2 different ways of setting "options", one via the script tag and one via the hash part of the UI. I'd like to make using the query string version to also work.
- How options work is currently inconsistent in how they work, IE some are "noXXX" while others are "XXX=false". 
- Extracting the options from query string is actually harder than it looks, in this process discovered bugs where if the setting was the first option or last option the system wouldn't work properly.
- Manipulating the query string is actually quite hard under Javascript, you need to remove the previous setting and add the new setting. Just appending a new value is fraught with danger and confusing.
- Settings come from multiple places, we have bootstrap.js settings, timing settings, extension settings. This is only going to get worse.
- All the above effect the creation of a GUI elements to control these settings.

While attempting to fix the above problems I ended up repeating very similar code in multiple locations, hence the refactoring towards having a settings framework.

mithro:
BTW Still very happy to discuss this change further on the whiteboard if you want.

PR #308, opened 2013-06-26
Use the tests array in the complete message.

mithro:
postMessage delivery order is not determanistic (IE you can get a result
message _after_ the complete message), so building the results array ourselves
is a bad idea. Plus why do that work when testharness already provides the list
for us!?

### DIFF
--- a/test/test-runner.html
+++ b/test/test-runner.html
@@ -455,40 +455,34 @@ window.addEventListener(
     var testname = evt.source.location.pathname.split('/').pop().replace('.', '-');
     var test = document.getElementById(testname);
 
-    switch(evt.data['type']) {
-    case 'start':
-      testResults[testname] = [];
-      break;
+    // We only respond to complete as postMessage doesn't guarantee order so
+    // result messages can come in after the complete message.
+    if (evt.data['type'] != 'complete')
+       return;
 
-    case 'result':
-      testResults[testname].push(evt.data['test']);
-      break;
+    var result = processResults(test, evt.data['tests']);
+    if (result && shouldPostResults) {
+      // Move from running to posting state
+      changeTestState(test, States.POSTING);
 
-    case 'complete':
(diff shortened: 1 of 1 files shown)


test/test-runner.html
```diff
@@ -487,7 +479,6 @@ <h1 id=status>Pending</h1>
         // Move directly to finished state
         changeTestState(test, States.FINISHED);
       }
-      break;
```
shans, line 490: Can you add a default: with an // Ignore comment to make it clear that ignoring the other messages is intentional?

LGTM after this change.

test/test-runner.html
```diff
@@ -456,16 +456,8 @@ <h1 id=status>Pending</h1>
     var test = document.getElementById(testname);
 
     switch(evt.data['type']) {
```
dstoc: Looks like we can remove the switch and use a simple if now.

shans:
Yeah, Doug's suggestion is better :)

mithro:
It's very likely that we'll want to add new message types, is it better to
change to an if now and then back to the switch or leave the switch?

dstoc:
> It's very likely that we'll want to add new message types, is it better to
> change to an if now and then back to the switch or leave the switch?

Yes, seems better to switch now.

mithro:
Sorry, I don't understand which method you're suggesting with your "better
to switch now" comment?

dstoc:
> Sorry, I don't understand which method you're suggesting with your "better
> to switch now" comment?

Better to switch to an 'if' now, and change back to a 'switch' statement if/when that's necessary in the future.

test/test-runner.html
mithro: Done!

test/test-runner.html
mithro, line 490: Done.

PR #310, opened 2013-06-26
Unpace path animation (by default)

alancutter:
I'm not sure we actually want this to land until we sort out the paced timing spec and implement it. I imagine most people would intend on their path animations being paced, this patch removes that behaviour with no easy way of getting it back via the API (you'd have to split your segments up into equal length portions).


alancutter:
Implemented paced timing for path animations ontop of this patch: https://github.com/web-animations/web-animations-js/pull/320

PR #312, opened 2013-06-26
Fixing the generator

mithro:
Depends on:
- https://github.com/web-animations/web-animations-js/pull/305
- https://github.com/web-animations/web-animations-js/pull/271

### DIFF
--- a/test/bootstrap.js
+++ b/test/bootstrap.js
@@ -77,6 +77,12 @@ function instrument(src) {
   var inst = window.__resources__[src] = new Instrumenter().instrumentSync(js, src);
 }
 
+// Can't use the setting framework to get this information because we are not
+// going to load any testharness stuff!
+if (/testharness=disable/.test(window.location.href)) {
+  loadScript('../../web-animations.js');
+  return;
+}
 
 loadScript('../testharness/testharness.js');
 loadCSS('../testharness/testharness.css');
--- a/test/bootstrap.js
+++ b/test/bootstrap.js
@@ -49,16 +49,10 @@ function loadScript(src, options) {
   }
 }
 
+window.bootstrap_loadScript = loadScript;
+
 function loadCSS(src) {
   document.write('<link rel="stylesheet" type="text/css" href="' + src + '">');
-}
-
-function hasFlag(flag) {
-  return thisScript && thisScript.getAttribute(flag) !== null;
-}
-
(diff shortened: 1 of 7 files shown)


steveblock:
I agree that this less intrusive approach looks better

lgtm

test/test-generator.html
```diff
@@ -184,20 +192,36 @@
...
 function printCheck(object, number, properties, time){
-  var newCheck = "check(document.querySelectorAll('" + object + "')[" + number +
-      "]," + properties + "," + time + ");";
+  var newCheck = "  at(" + time + ", function() {\n" +
+     '    assert_styles("' + object + '", [\n'
+
+  // If all the properties are the same for every object, just pass the list.
+  var allsame = properties.every(function(x) { return x === properties[0] });
+  if (allsame) {
+    newCheck = newCheck.substring(0, newCheck.length-2);
```
alancutter, line 201: You could avoid this string truncation by adding the "[\n" in the else instead. This would be cleaner IMO.

mithro, line 201: Fixed!

test/bootstrap.js
```diff
@@ -83,22 +77,26 @@ function instrument(src) {
   var inst = window.__resources__[src] = new Instrumenter().instrumentSync(js, src);
 }
 
+// Can't use the setting framework to get this information because we are not
+// going to load any testharness stuff!
+if (/testharness=disable/.test(window.location.href)) {
+  loadScript('../../web-animations.js');
```
alancutter, line 83: I think you should be loading ../testharness_stub.js here.

mithro, line 83: Fixed!

alancutter:
Some of our tests use scheduled events to check non-style values or alter the animation model, with the test harness disabled these actions are hard to time right and generate results for. This patch doesn't cover these cases.

As much as I'd like to have the generator working again I think this may be the wrong solution and merging it now will make fixing the issue more difficult than it needs to be. The generator should be using the testharness to control time just like the regular tests do.

mithro:
This is what my first attempt at fixing the generate did, see
https://github.com/mithro/web-animations-js/tree/fixing-generator

Maybe we need to merge the two methods?

mithro:
Will refactor and resubmit as a pull request...

PR #316, opened 2013-06-27
Fix MediaReference after switch to new Timing interface

steveblock:
The test runs fine in Chrome, but seems flaky with run-tests.py - investigating

### DIFF
--- a/demos/curtains.html
+++ b/demos/curtains.html
@@ -57,13 +57,13 @@ var duration = 2.0;
 video.addEventListener("loadeddata", function() {
   document.timeline.play(new SeqGroup([
     new ParGroup([
-      new Animation(leftCurtain, {width: "0px"}, {duration: duration}),
-      new Animation(rightCurtain, {width: "0px"}, {duration: duration}),
+      new Animation(leftCurtain, {width: "0px"}, {iterationDuration: duration}),
+      new Animation(rightCurtain, {width: "0px"}, {iterationDuration: duration}),
     ]),
     new MediaReference(video, {startDelay: -duration}),
     new ParGroup([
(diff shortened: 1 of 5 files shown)


steveblock:
It looks like the cause of the test flakiness is a bug in Chrome which means that MediaElement.seekable isn't set properly when the video is served with a 200 response code, rather than a 206 response code. This is a problem because the test harness' local server uses a 200 response code. See https://code.google.com/p/chromium/issues/detail?id=256508

Until this is fixed, I think there are a few options ...
1. Load the videos via XHR and a data URL. This requires https://github.com/web-animations/web-animations-js/issues/321 to be fixed first.
2. Configure the test harness' local server to always use a 206 response code for videos. This has the downside that it people run tests manually using a python server, it won't be similarly configured.
3. Load the test videos from a remote server which always uses a 206 response code. We could use a github gh-pages branch for this.

I'll go for option 3.

steveblock:
I've updated my commits to use approach 3. I added the video files to the gh-pages branch of my fork of the repository and have modified the tests to fetch them from the corresponding github URL. If this looks OK, we can switch to using a gh-pages branch in the main repository.

> 1. Load the videos via XHR and a data URL. This requires #321 to be fixed first.
>    Note that this is true even if you use a synchronous XHR to get the data URL. Loading a video is always asynchronous, even from a data URL, and the page load event doesn't wait for content that's added with document.write().

steveblock:
It turns out that there's no way to force video elements to load before page load, so these tests are always going to be flaky until https://github.com/web-animations/web-animations-js/issues/321 is fixed.

I've updated my patch to retain the tests, but remove test-media.html from testcases.js.

PR #317, opened 2013-06-27
Remove bad change to test_without_at

alancutter:


mithro:
@alancutter The desc is currently hoisted from the outer scope right?

alancutter:
The desc is being hoisted from the outer scope, this is desired behaviour in many tests.
This is only not desired when multiple at tests use the same t value within the same timing_test function.
The original patch was to make the desc optional for at and went a little overboard.

For test_without_at there's no reason to override the original desc so it should be hoisted into the function that calls testharness_test. This patch brings it back to that state.

PR #318, opened 2013-06-27
Test desc another attempt fix

mithro:
Can you guys have a look, hopefully we've gotten it right this time.
@alancutter
@steveblock


test/testharness_timing.js
```diff
@@ -599,12 +599,16 @@
...
-        var at = function(seconds, f, atDesc)
+        var at = function(seconds, f, desc_at)
         {
             assert_true(typeof seconds == "number", "at's first argument shoud be a number.");
             assert_true(typeof f == "function", "at's second argument should be a function.");
 
-            var t = async_test(atDesc ? atDesc : (desc + " at t=" + seconds + "s"));
+            // Provide a default value..
+            if (typeof desc_at == "undefined" || desc_at == null || desc_at.length == 0)
+              desc_at = "Unnamed";
```
alancutter: I think the appropriate thing to do when no description is given in the at is to use the desc given to timing_test.

mithro: Actually, that is probably reasonable. Fixing....

mithro:
Fixed. @alancutter please merge!

PR #319, opened 2013-06-27
Add property-indexed keyframe syntax

dstoc:


dstoc:
On second thoughts, the sorting is kind of bogus. Let me think about that some more.

web-animations.js
```diff
@@ -1554,6 +1553,142 @@ var normalizeKeyframeDictionary = function(properties) {
   return result;
 };
 
+var parseKeyframeInputObject = function(keyframeInput) {
```
shans: It took me ages to realize that this was only called if keyframeInput wasn't already an array. Either call this arrayFromKeyframeInput, or move the array check to this function.

Or do both. This isn't really parsing :)

web-animations.js
```diff
@@ -1554,6 +1553,142 @@ var normalizeKeyframeDictionary = function(properties) {
...
+    if (Array.isArray(keyframeInput[k])) {
+      containsArray = true;
+    }
+  }
+  if (!containsArray) {
+    return [keyframeInput];
+  }
+  // Otherwise must be property-indexed.
+  var result = [];
+  for (var k in keyframeInput) {
```
shans: Use a more descriptive name than k. Each of these is a property, right?

web-animations.js
```diff
@@ -1554,6 +1553,142 @@ var normalizeKeyframeDictionary = function(properties) {
...
+  for (var i = length - 1; i >= 0; i--) {
+    var offset = keyframes[i].offset;
+    if (isDefinedAndNotNull(offset)) {
+      if (offset <= 1) {
+        break;
+      } else {
+        count = length - i;
+      }
+    }
+  }
```
shans, line 1869: Use filter for the above two steps:

keyframes = keyframes.filter(function(frame) { return !isDefinedAndNotNull(offset) || (frame.offset >= 0 && frame.offset <= 1); });

web-animations.js
```diff
@@ -1554,6 +1553,142 @@ var normalizeKeyframeDictionary = function(properties) {
...
+    if (i > nextOffsetIndex) {
+      nextOffsetIndex = i;
+      while (!isDefinedAndNotNull(keyframes[nextOffsetIndex].offset)) {
+        nextOffsetIndex++;
+      }
+    }
+    var a = keyframes[lastOffsetIndex].offset;
+    var b = keyframes[nextOffsetIndex].offset;
+    var n = nextOffsetIndex - lastOffsetIndex - 1;
+    keyframes[i].offset = a + (b - a) * i / (n + 1);
```
shans: You need to do this in a loop as you may have consumed multiple 'i's.

shans: Oh, that's wrong. But I think it would actually be better to use a loop here and skip i forward, rather than checking keyframe.offset each iteration.

web-animations.js
dstoc: Renamed.

web-animations.js
dstoc: Done.

web-animations.js
dstoc, line 1869: I'm only moving this code. Let's clean it up in a separate change.

web-animations.js
dstoc: Ditto.

dstoc:
@birtles see https://github.com/dstockwell/web-animations-js/blob/b3a9c28595b806598bd6a97770975f0d4146a463/test/testcases/unit-test-property-indexed-keyframes.html

This is the approach I described the last time we discussed how we might normalize property-indexed keyframes.

birtles:
I thought you were going to write a proposal for this so we could work out if was a good idea or not? Or is this part of the feasibility study?

dstoc:
It was:

> ACTION: Doug to work out what this would look like from an API point of view

I can write it up to discuss on Friday.

PR #325, opened 2013-06-28
Fixing issue #264 by not loading testharness_timing.js in unittests.

mithro:
It was descided that unittests should not be time dependent and should not use at/timing_test functionality.


mithro:
See https://github.com/web-animations/web-animations-js/issues/264

Issue #273, opened 2013-06-22
update-testcases should update the test cases


dstoc:
Why isn't that the default with a --no-update flag if you don't actually want to?

This has bitten me a few times since it prints the same output in both cases. If you really want --no-update to be the default, perhaps it could print out a message, ideally in red.

PR #330, opened 2013-06-30
update-testcases.py should update the list of testcases by default

steveblock:
Added a --dry-run option to print the diff that would otherwise be applied.

This fixes https://github.com/web-animations/web-animations-js/issues/273

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ install:
   - ./.travis-setup.sh
 
 script:
-  - python ./test/update-testcases.py || (echo "Please run update-testcases!"; false)
+  - python ./test/update-testcases.py --dry-run || (echo "Please run update-testcases!"; false)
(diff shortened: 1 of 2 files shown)


dstoc:
Does .travis.yml need to be updated to pass --dry-run ?

mithro:
Yes .travis.yml needs --dry-run if this change is committed.

The reason I made it not change by default is that I didn't want someone to change the testcases without actually deliberately meaning to.

steveblock:
> The reason I made it not change by default is that I didn't want someone to change the testcases without
> actually deliberately meaning to.
> Makes sense, but I think that's unlikely given that the name of the command and the fact that they still have to commit the change and get it reviewed. I've updated the patch to always print the diff to make things even more explicit.
> 
> Yes .travis.yml needs --dry-run if this change is committed.
> Done

PR #335, opened 2013-07-01
Implement accumulate for KeyframeAnimationEffect

steveblock:


dstoc:
lgtm

steveblock:
Snahe, could you take a look before we merge this to check you agree on the logic about when and how to accumulate? The spec text is rather vague, as it talks about the sum of CompositableValues - http://dev.w3.org/fxtf/web-animations/#accumulating-animation-values

web-animations.js
```diff
@@ -1646,36 +1646,34 @@ KeyframeAnimationEffect.prototype = createObject(AnimationEffect.prototype, {
...
-      }
+  _sampleForProperty: function(timeFraction, currentIteration, property) {
+    var frames = this._propertySpecificKeyframes(property);
+    var unaccumulatedValue =
+        this._unaccumulatedValueForProperty(frames, timeFraction, property);
+    // We can only accumulate if this iteration is strictly positive and if all
+    // keyframes use the same composite operation. As an optimisation, if that
+    // common composite opertation is 'replace', we can skip accumulating.
+    if (this.accumulate === 'none' || currentIteration <= 0 ||
+        !this._allKeyframesUseCompositeAdd(frames)) {
```
shans: This doesn't seem to do what the comment says.

this._allKeyframesUseCompositeAdd(..) is true if every frame is 'add', right? So !this._allKeyframesUseCompositeAdd(..) just means at least one frame uses 'replace'.

shans: Actually, this is checking that all frames use the same composite operation.

I think it's actually the comment that is wrong. I don't think we can skip accumulating if the common composite operation is 'replace'. The code doesn't, and I think that's correct.

web-animations.js
```diff
@@ -3328,6 +3362,38 @@ BlendedCompositableValue.prototype =
...
+    // better handle large numbers of iterations.
+    // TODO: We only use this type when both accumulatingValue and topmostValue
+    // are instances of AddReplaceCompositableValue with composite='add'.
+    // Consider optimising by generating a single AddReplaceCompositableValue
+    // rather than making repeated calls to compositeOnto().
+    var result = underlyingValue;
+    for (var i = 0; i < this.accumulationCount; i++) {
+      result = this.accumulatingValue.compositeOnto(property, result);
+    }
+    return this.topmostValue.compositeOnto(property, result);
```
shans: compositeOnto is wrong here, I'm pretty sure. We need to always sum the values, even if the composite operation is 'replace':

'The intermediate animation value produced by the animation effect for a given property is the sum of the unaccumulated animation value for the given time fraction and the final intermediate animation value of the previous iteration'

So the 'sum' applies whether or not the operation is 'add'.

shans:
I think the replace behaviour is incorrect, but the general approach LGTM.

steveblock:
Fixed behaviour for replace composite. PTAL.

web-animations.js
```diff
@@ -1655,35 +1655,57 @@ KeyframeAnimationEffect.prototype = createObject(AnimationEffect.prototype, {
...
+          this._getAccumulatingValue(frames, property), currentIteration);
+    }
+
+    return unaccumulatedValue;
+  },
+  _getAccumulatingValue: function(frames, property) {
+    console.assert(this._allKeyframesUseSameCompositeOperation(frames),
+        'Accumulation only valid if all frames use same composite operation');
+
+    // This is a BlendedCompositableValue, though becuase the offset is 1.0, we
```
shans: becuase -> because

shans:
LGTM.

This is really quite complex behaviour for a feature that seems destined to be unused except for SVG support. I wonder if there's some way to push it out of the spec and into the SVG binding at some point?

PR #336, opened 2013-07-03
Implemented pauseAnimationsForTesting + test.

rjwright:


steveblock:
A general comment about commits ...

Deciding how fine-grained your commits should be is a bit of a balancing act. There's some personal preference involved, but in general, having too few makes things hard to review and having too many adds clutter which makes it harder for anybody doing code archaeology in the future.

I think this pull request is in the latter category - this could be a single commit. Particularly as the comment in the first commit is then negated in that of the second commit!

There's a similar trade-off with whether changes made in response to review comments should be merged into the existing commits or made as subsequent commits. It probably depends on how many logically independent commits make up your pull request and which of them need changes.

Let me know if you need a hand with rebasing!

rjwright:
Yeah. I know my commits have been small and not well organised. Mostly it's
because I'm hopeless with git. It's only dawning on me now that reviewers
see the commits - that they're not just part of personal workflow.

I need a hand getting my code up to date :(

steveblock:
OK, why don't you carry on as you are for now, and we can do some git surgery tomorrow morning.

rjwright:
PTAL, thanks.

web-animations.js
```diff
@@ -209,8 +209,15 @@ Timeline.prototype = {
...
   },
+  _pauseAnimationsForTesting: function(pauseAt) {
+    PLAYERS.forEach(function(player) {
+      player.paused = true;
+      player.currentTime = pauseAt;
+    });
+  },
 };
 
+
```
dstoc: There's an extra line here.

test/testcases/test-pause-for-testing.html
```diff
@@ -0,0 +1,59 @@
...
+  document.timeline.play(downBlue);
+  document.timeline._pauseAnimationsForTesting(2);
+
+  timing_test(function() {
+    at(0, function() {
+      assert_styles('#a', {'top':'300px'})
+      assert_styles('#b', {'top':'150px'})
+    });
+  },
+  "Pause all aimations in the document and seek them to a given time." +
```
dstoc: I think this is missing a space, it ends up with something like "... time.Test that ..."

dstoc: Best to be concise, maybe something like "Pausing all animations should work when players have different start delays"

test/testcases/test-pause-for-testing.html
```diff
@@ -0,0 +1,59 @@
...
+    border-radius: 50%;
+    }
+</style>
+<div class="anim" id="a" style="left: 0px; background-color: lightsteelblue"></div>
+<div class="anim" id="b" style="left: 450px; background-color: red"></div>
+<script src="../bootstrap.js"></script>
+<script>
+  "use strict";
+  var downRed = new ParGroup(
+    [new Animation(document.querySelector("#a"), {top: "600px"}, 4)]);
```
dstoc: When wrapping you should indent at least 4 spaces.

test/testcases/test-pause-for-testing.html
```diff
@@ -0,0 +1,59 @@
...
+    height: 50px;
+    top: 0px;
+    border-radius: 50%;
+    }
+</style>
+<div class="anim" id="a" style="left: 0px; background-color: lightsteelblue"></div>
+<div class="anim" id="b" style="left: 450px; background-color: red"></div>
+<script src="../bootstrap.js"></script>
+<script>
+  "use strict";
```
dstoc: Prefer single quotes over double quotes.

test/testcases/test-pause-for-testing.html
```diff
@@ -0,0 +1,59 @@
...
+    at(0, function() {
+      assert_styles('#a', {'top':'300px'})
+      assert_styles('#b', {'top':'150px'})
+    });
+  },
+  "Pause all aimations in the document and seek them to a given time." +
+  "Test that it works when players have different start delays.")
+
+  timing_test(function() {
+    at(12, function() {
```
dstoc: This should be part of the timing_test above.

test/testcases/test-pause-for-testing.html
rjwright: Everywhere, e.g. in test assertions?

dstoc: I think the only places you need to do this are here and on lines 36 and 37.

The way you are indenting the anonymous functions is fine.

rjwright: Sweet. It does look nicer that way :)

PR #341, opened 2013-07-07
Fix keyframe distribution

steveblock:


web-animations.js
```diff
@@ -1894,7 +1894,10 @@ KeyframeAnimationEffect.prototype = createObject(AnimationEffect.prototype, {
       var a = this._cachedDistributedKeyframes[lastOffsetIndex].offset;
       var b = this._cachedDistributedKeyframes[nextOffsetIndex].offset;
       var n = nextOffsetIndex - lastOffsetIndex - 1;
-      this._cachedDistributedKeyframes[i].offset = a + (b - a) * i / (n + 1);
+      console.assert(n > 0);
+      var j = i - lastOffsetIndex;
```
dstoc: It's not clear what j is. I think there are probably better variable names for a, b, n and j. i is OK :)

steveblock: These are the variable names used in the spec, but I'll update them to be more descriptive.

PR #342, opened 2013-07-07
Add a test that applying accumulation takes immediate effect

steveblock:


test/testcases/test-keyframe-animation-effect.html
```diff
@@ -60,8 +61,16 @@
   assert_equals(effect.accumulate, "sum");
 }, "KeyframeAnimationEffect.accumulate should only accept valid values");
 
-// TODO: Test that setting setting KeyframeAnimationEffect.accumulate causes
-// an immediate update to its target. Need to implement accumulation first!
+timing_test(function() {
+  var animation = new Animation(document.getElementById("anim0"), keyframes,
```
dstoc, line 65: I think element.animate would be simpler here

PR #345, opened 2013-07-09
Clean up the fooType classes used for adding and interpolating property values

steveblock:


dstoc:
lgtm, but @alancutter should take a look

web-animations.js
```diff
@@ -2148,34 +2148,64 @@ var interpArray = function(from, to, f, type) {
...
+    value = clamp(value, 100, 900);
+    if (value === 400) {
+      return 'normal';
+    }
+    if (value === 700) {
+      return 'bold';
+    }
+    return String(value);
   },
   fromCssValue: function(value) {
```
alancutter, line 2204: fromCssValue() should probably check for valid input like the other types now.

web-animations.js
```diff
@@ -2315,7 +2359,18 @@ var rectangleType = {
...
+        top: percentLengthType.fromCssValue('auto'),
+        right: percentLengthType.fromCssValue('auto'),
+        bottom: percentLengthType.fromCssValue('auto'),
+        left: percentLengthType.fromCssValue('auto'),
+      };
+    }
     var match = rectangleRE.exec(value);
+    if (!match) {
+      return undefined;
+    }
```
alancutter, line 2377: undefined should be returned if any of the following calls to percentLengthType.fromCssValue() return undefined as well.

alancutter:
visibilityType.fromCssValue() and colorType.fromCssValue() should check for valid input as well.

alancutter:
This is a much needed change! We were quite lacking in this area of input validation.

alancutter:
Created pull request for changes to this branch: https://github.com/steveblock/web-animations-js/pull/1

PR #346, opened 2013-07-09
Fix matrix transform interpolation

alancutter:


steveblock:
Can you add a test?

alancutter:
I thought the generator was broken but it seems to be working for simple test cases.
Added test.

PR #348, opened 2013-07-10
Fixing the testharness to work when there are no tests.

### LINKED ISSUE #347: Testharness breaks when there are no tests to run

mithro:
Fixes #347 reported by Alan.

### DIFF
--- a/test/testharness_timing.js
+++ b/test/testharness_timing.js
@@ -345,7 +345,6 @@
         }
     };
 
-
     /**
      * Return the current time in milliseconds.
      */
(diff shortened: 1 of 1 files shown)


test/testharness_timing.js
```diff
@@ -548,8 +568,14 @@
 
         var t = ts - raf_t0;
 
+        var endTime = testharness_timeline.endTime_;
+        // If we have no events paste t=0, endTime is going to be zero. Instead
```
dstoc, line 551: Can't we just do this when endTime is 0?

mithro, line 551: Yes, you are right. That makes this patch much similar.

mithro:
PTAL.

PR #350, opened 2013-07-10
Style assertion fix for properties dependent on other properties

alancutter:
The computed value of certain CSS properties can be affected by others.
This patch copies the style properties from the test element onto the reference element before individual properties are compared.

### DIFF
--- a/test/testharness_extensions.js
+++ b/test/testharness_extensions.js
@@ -171,6 +171,12 @@
         var reference_element = (object.namespaceURI == svg_namespace_uri) ?
             document.createElementNS(svg_namespace_uri, object.nodeName) :
             document.createElement(object.nodeName);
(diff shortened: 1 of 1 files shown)


test/testharness_extensions.js
```diff
@@ -206,7 +206,7 @@
                   reference_element.style[output_prop_name] = prop_value;
 
                   var current_style = getComputedStyle(object, null);
-                  var target_style = getComputedStyle(reference_element, null);
```
dstoc, line 209: Isn't this the entire point of using the reference element?

dstoc:
Not lgtm. I would rather keep the conversion/normalization behavior.

Perhaps a better fix would be to copy all of the other style properties from the element to the reference element.

alancutter:
Updated fix to copy style properties across.

PR #354, opened 2013-07-12
Reduce recalc/layout during the compositing of animated values.

dstoc:
Previously we would cause a recalc and layout for each animated
target/property pair, for every frame. With this change only one recalc/layout
is required per frame.


dstoc:
lgtm was here: https://github.com/dstockwell/web-animations-js/commit/9d122c234d2a113e8c6973cf0f4208184c91e63e#commitcomment-3619378

PR #357, opened 2013-07-12
Avoid generating calc() expressions when just adding a 0 value.

dstoc:


web-animations.js
```diff
@@ -2276,9 +2276,11 @@ var percentLengthType = {
...
-        s = features.calcFunction + '(' + s + ' + ' + value[item] + item + ')';
-        single_value = false;
-      } else {
+        if (value[item] != 0) {
+          s = features.calcFunction + '(' + s + ' + ' + value[item] + item + ')';
+          single_value = false;
+        }
+      } else if (value[item] != 0) {
         s = s.substring(0, s.length - 1) + ' + ' + value[item] + item + ')';
       }
```
alancutter, line 2285: This won't cull off 0 values when they are the first value.
Example: percentLengthType.toCssValue({ '%': 0, 'px': 10}) === 'calc(0% + 10px)'

dstoc, line 2285: Yep, this was just a small optimization. We could do more work here if it's worthwhile.

PR #359, opened 2013-07-12
Optimize player sorting.

dstoc:
- Sort by sequence number per spec.
- Don't require a stable sort.
- Track and invalidate the sorted state.

### DIFF
--- a/web-animations.js
+++ b/web-animations.js
@@ -219,12 +219,15 @@ Timeline.prototype = {
 
 // TODO: Remove dead Players from here?
 var PLAYERS = [];
+var sortedPlayers;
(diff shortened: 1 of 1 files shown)


web-animations.js
```diff
@@ -1604,7 +1609,7 @@ var normalizeKeyframeDictionary = function(properties) {
   }
   // TODO: Remove prefixed properties if the unprefixed version is also
   // supported and present.
-  animationProperties = animationProperties.sort();
+  animationProperties = animationProperties.sort(playerSortFunction);
```
dstoc, line 1612: Oops.

PR #360, opened 2013-07-13
Implement positionListType for animating CSS property background-position

alancutter:


test/disabled-testcases
```diff
@@ -1,5 +1,10 @@
 # Lots of problems
 test-change-playback-rate.html
+
 # Requires the test harness to allow tests to be added after page load. See
 # https://github.com/web-animations/web-animations-js/issues/321
 test-media.html
+
+# Blink doesn't output calc background-position values properly
+# crbug.com/260097
+test-position-list-type.html
```
dstoc: Can't we enable it with the generated checks?

alancutter: The generated checks would be bogus in many cases. If this test were to be enabled I would enable it in browsers that can output computed background-position values properly (Firefox) though that'd be another patch for run_tests.py.

dstoc: I think it's fine for them to be bogus. They are just auto-generated expectations. The purpose is not necessarily to verify that the behavior is correct, but to avoid breaking the code in the future. I'd prefer we have coverage for the code rather than avoid checking in bogus expectations.

alancutter: Generating the test results is a bit annoying due to this Blink bug and the way it interacts with assert_styles.

The generated styles will contain values like '10% ' that should have been '10% calc(...)'.
The reference's computed style becomes '10% 50%'. This is then compared with the actual computed value which is read as '10% ' and fails the check.

There doesn't seem to be any trivial ways to get around this without altering the value normalisation that assert_styles performs.

PR #364, opened 2013-07-18
Reworking test-runner and cleaning up the realtime runner stuff a little.

mithro:
(This cleanup make the Firefox fixes easier.)

### DIFF
--- a/test/bootstrap.js
+++ b/test/bootstrap.js
@@ -539,6 +539,84 @@ TestTimelineGroup.prototype.draw = function(container, endTime) {
   this.info.style.display = 'none';
 };
 
+
+
+/**
+ * Moves the testharness_timeline in "real time".
+ * (IE 1 test second takes 1 real second).
+ *
+ * @constructor
+ */
+function RealtimeRunner(timeline) {
+  this.timeline = timeline;
+
+  // Capture the real requestAnimationFrame so we can run in 'real time' mode
+  // rather than as fast as possible.
+  var nativeRequestAnimationFrame = window.requestAnimationFrame || window.mozRequestAnimationFrame;
+  this.boundRequestAnimationFrame = function(f) {
+    nativeRequestAnimationFrame(f.bind(this))
+  };
+  this.now = window.Date.now;
+
+  this.zeroTime = null;       // Time the page loaded
+  this.pauseStartTime = null; // Time at which we paused raf
+  this.timeDrift = 0;         // Amount we have been stopped for
+}
+
+/**
+ * Callback called from nativeRequestAnimationFrame.
+ *
+ * @private
+ * @param {number} timestamp The current time for the animation frame
+ *   (in millis).
+ */
+RealtimeRunner.prototype.animationFrame_ = function(timestamp) {
+  if (this.zeroTime === null) {
+    this.zeroTime = timestamp;
+  }
+
+  // Are we paused? Stop calling requestAnimationFrame.
+  if (this.pauseStartTime != null) {
+    return;
+  }
+
+  var virtualAnimationTime = timestamp - this.zeroTime - this.timeDrift;
+  var endTime = this.timeline.endTime_;
+  // If we have no events paste t=0, endTime is going to be zero. Instead
+  // make the test run for 2 minutes.
+  if (endTime == 0) {
+    endTime = 120e3;
+  }
+
+  // Do we still have time to go?
+  if (virtualAnimationTime < endTime) {
+    this.timeline.setTime(virtualAnimationTime);
+    this.boundRequestAnimationFrame(this.animationFrame_);
+
+  } else {
+    // Have we gone past endTime_? Force the harness to its endTime_.
+
+    this.timeline.setTime(endTime);
+    // Don't continue to raf
+  }
+};
+
+RealtimeRunner.prototype.start = function() {
+  if (this.pauseStartTime != null) {
+    this.timeDrift += (this.now() - this.pauseStartTime);
+    this.pauseStartTime = null;
+  }
+  this.boundRequestAnimationFrame(this.animationFrame_);
+};
+
+RealtimeRunner.prototype.pause = function() {
+  if (this.pauseStartTime != null) {
+    return;
+  }
+  this.pauseStartTime = this.now();
+};
+
+
 /**
  * Class for storing events that happen during at given times (such as
  * animation checks, or setTimeout).
@@ -563,6 +641,8 @@ function TestTimeline(everyFrame) {
   this.schedule(function() {}, 0);
 
   this.reset();
+
+  this.runner_ = new RealtimeRunner(this);
 }
 
 /**
@@ -601,10 +681,10 @@ TestTimeline.prototype.createGUI = function(body) {
   this.control.id = 'control';
   this.control.onclick = function() {
     if (this.control.innerText == 'Go!') {
-      testharness_raf_start();
+      this.runner_.start();
       this.control.innerText = 'Pause';
     } else {
-      testharness_raf_pause();
+      this.runner_.pause();
       this.control.innerText = 'Go!';
     }
   }.bind(this);
@@ -850,65 +930,6 @@ TestTimeline.prototype.autorun = function() {
   this.toNextEvent();
 };
 
-// Capture the real requestAnimationFrame so we can run in 'real time' mode
-// rather than as fast as possible.
-var raf = window.requestAnimationFrame;
-var raf_now = Date.now;
-
-var raf_t0 = null;    // Time the page loaded
-var raf_pauseStartTime = null; // Time at which we paused raf
-var raf_timeDrift = 0;    // Amount we have been stopped for.
-function testharness_raf(ts) {
-  if (raf_t0 === null) {
-    raf_t0 = ts;
-  }
-
-  // Are we paused? Stop rafing.
-  if (raf_pauseStartTime != null) {
-    return;
-  }
-
-  var t = ts - raf_t0 - raf_timeDrift;
-
-  var endTime = testharness_timeline.endTime_;
-  // If we have no events paste t=0, endTime is going to be zero. Instead
-  // make the test run for 2 minutes.
-  if (endTime == 0) {
-    endTime = 120e3;
-  }
-
-  // Do we still have time to go?
-  if (t < endTime) {
-    testharness_timeline.setTime(t);
-    raf(testharness_raf);
-
-  } else {
-    // Have we gone past endTime_? Force the harness to its endTime_.
-
-    testharness_timeline.setTime(testharness_timeline.endTime_);
-    // Don't continue to raf
-  }
-
-  // FIXME: When reset is called, we need to clear raf_t0
-}
-
-function testharness_raf_start() {
-  if (raf_pauseStartTime == null) {
-    return;
-  }
-
-  raf_timeDrift += (raf_now() - raf_pauseStartTime);
-  raf_pauseStartTime = null;
-
-  raf(testharness_raf);
-}
-
-function testharness_raf_pause() {
-  if (raf_pauseStartTime != null) {
-    return;
-  }
-  raf_pauseStartTime = raf_now();
-}
 
 
 function testharness_timeline_setup() {
@@ -935,7 +956,8 @@ function testharness_timeline_setup() {
 
   } else if('#explore' == window.location.hash ||
         window.location.hash.length == 0) {
-    raf(testharness_raf);
+
+    setTimeout(testharness_timeline.runner_.start.bind(testharness_timeline.runner_), 1);
   } else {
     alert('Unknown start mode.');
   }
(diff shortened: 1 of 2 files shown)


test/bootstrap.js
```diff
@@ -535,6 +535,76 @@ TestTimelineGroup.prototype.draw = function(container, endTime) {
...
 };
 
+
+
+function RealtimeRunner(timeline) {
+  this.timeline = timeline;
+
+  // Capture the real requestAnimationFrame so we can run in 'real time' mode
+  // rather than as fast as possible.
+  var ourraf = window.requestAnimationFrame || window.mozRequestAnimationFrame;
```
dstoc: ourraf -> localRaf or nativeRaf or even nativeRequestAnimationFrame

test/bootstrap.js
```diff
@@ -535,6 +535,76 @@ TestTimelineGroup.prototype.draw = function(container, endTime) {
...
+  this.raf = function(f) { ourraf(f.bind(this)) };
+  this.now = window.Date.now;
```
dstoc, line 559: nit: we should really be using performance.now

test/bootstrap.js
```diff
@@ -535,6 +535,76 @@ TestTimelineGroup.prototype.draw = function(container, endTime) {
...
+
+  this.t0 = null;             // Time the page loaded
```
dstoc: z0 -> timeZero

test/bootstrap.js
```diff
@@ -535,6 +535,76 @@ TestTimelineGroup.prototype.draw = function(container, endTime) {
...
+  this.pauseStartTime = null; // Time at which we paused raf
+  this.timeDrift = 0;         // Amount we have been stopped for
+}
+
+/**
+ * Class for storing events that happen during at given times (such as
```
dstoc: 'Stores events ...' we know it's a class.

dstoc: This comment and @constructor should be on the definition of RealTimeRunner above.

test/bootstrap.js
```diff
@@ -535,6 +535,76 @@ TestTimelineGroup.prototype.draw = function(container, endTime) {
...
+ * animation checks, or setTimeout).
+ *
+ * @constructor
+ */
+RealtimeRunner.prototype.animationFrame = function(ts) {
```
dstoc: what is ts? needs a better name

test/bootstrap.js
```diff
@@ -535,6 +535,76 @@ TestTimelineGroup.prototype.draw = function(container, endTime) {
...
+  if (this.t0 === null) {
+    this.t0 = ts;
+  }
+
+  // Are we paused? Stop rafing.
+  if (this.pauseStartTime != null) {
+    return;
+  }
+
+  var t = ts - this.t0 - this.timeDrift;
```
dstoc: t = currentTime ?

test/bootstrap.js
```diff
@@ -535,6 +535,76 @@ TestTimelineGroup.prototype.draw = function(container, endTime) {
...
+  // make the test run for 2 minutes.
+  if (endTime == 0) {
+    endTime = 120e3;
+  }
+
+  // Do we still have time to go?
+  if (t < endTime) {
+    this.timeline.setTime(t);
+    this.raf(this.animationFrame);
+
```
dstoc, line 595: extra blank line here

test/bootstrap.js
```diff
@@ -535,6 +535,76 @@ TestTimelineGroup.prototype.draw = function(container, endTime) {
...
+ * animation checks, or setTimeout).
+ *
+ * @constructor
+ */
+RealtimeRunner.prototype.animationFrame = function(ts) {
+  if (this.t0 === null) {
+    this.t0 = ts;
+  }
+
+  // Are we paused? Stop rafing.
```
dstoc: We should be able to write the code in such a way that these trivial comments are not necessary.

test/test-runner.html
```diff
@@ -101,6 +99,21 @@ <h1 id=status>Pending</h1>
   window.__resources__ = {original: {}};
 }
 
+window.addEventListener('load', function() {
+  document.querySelector('#spacer'
+    ).style.height = getComputedStyle(document.querySelector('body')).height;
```
dstoc: wrapping is odd
try:

```
document.querySelector('#spacer').style.height =
    getComputedStyle(document.querySelector('body')).height;
```

test/test-runner.html
```diff
@@ -101,6 +99,21 @@ <h1 id=status>Pending</h1>
...
+});
+
+function trueOffsetTop(element) {
+  var t = 0;
+  if (element.offsetParent) {
+    do {
+      t += element.offsetTop;
+    } while (element = element.offsetParent);
+  return [t];
```
dstoc: indent

dstoc: why does this return an array?

test/test-runner.html
```diff
@@ -267,25 +256,54 @@ <h1 id=status>Pending</h1>
...
-/* @type {Array.<Element>} */ var runningTests = [];
-/* @type {Array.<Element>} */ var postingTests = [];
-/* @type {Array.<Element>} */ var finishedTests = [];
 
 /**
  * Enum for state values.
  * @enum {number}
  */
 var States = {
-  LOADING: 1,    /* Test which is being loaded. */
```
dstoc, line 281: do we rely on these integers? or can we just use string values and remove the mapping function below?

test/test-runner.html
```diff
@@ -297,100 +315,154 @@ <h1 id=status>Pending</h1>
...
+  if (newState == States.POSTING && skipPosting) {
+    newState = States.POSTED;
   }
-  i = runningTests.indexOf(test);
-  if (i >= 0) {
-    runningTests.splice(i, 1);
-    testFinished(test);
+
+  var oldState = test.state;
+  if (typeof oldState != "undefined") {
```
dstoc: prefer single quotes

test/test-runner.html
```diff
@@ -297,100 +315,154 @@ <h1 id=status>Pending</h1>
...
   switch(newState) {
   case States.LOADING:
-    loadingTests.unshift(test);
     runTestsIfLoaded();
     break;
-  case States.UNSTARTED:
-    unstartedTests.unshift(test);
+
+  case States.LOADED:
+    if (!(oldState == States.LOADING)) {
```
dstoc: if (oldState != States.LOADING)

test/test-runner.html
```diff
@@ -297,100 +315,154 @@ <h1 id=status>Pending</h1>
...
+    }
     break;
+
   case States.RUNNING:
-    testStarted(test);
-    runningTests.unshift(test);
-    break;
-  case States.POSTING:
-    postingTests.unshift(test);
+    if (!(oldState == States.LOADED)) {
```
dstoc: ditto

test/test-runner.html
```diff
@@ -297,100 +315,154 @@ <h1 id=status>Pending</h1>
...
+      throw "Unexpected state change! " + stateToString(oldState) + " changing to RUNNING";
+    }
+
+    test.start = now();
+    var testWindow = test.querySelector('iframe').contentWindow;
+    var msg = {'type': 'start', 'url': testWindow.location.href + ''};
```
dstoc: no quotes around the property names

test/test-runner.html
```diff
@@ -297,100 +315,154 @@ <h1 id=status>Pending</h1>
...
+
   case States.FINISHED:
-  case States.TIMEOUT:
-    finishedTests.unshift(test);
+    var timingInfo = test.querySelector('.timing');
+    if (test.start) {
+      test.finished = now();
+      timingInfo.textContent = (test.finished - test.start).toFixed(2) + 'ms';
+    } else {
+      timingInfo.textContent = "(Load only)";
```
dstoc: prefer single quotes

test/test-runner.html
```diff
@@ -297,100 +315,154 @@ <h1 id=status>Pending</h1>
...
+    }
+
+    // If a test doesn't use any timing stuff it could come from the LOADING or
+    // LOADED state into the FINISHED state bypassing the RUNNING state.
+    if (!(oldState == States.LOADING || oldState == States.LOADED || oldState == States.RUNNING)) {
+      throw "Unexpected state change! " + stateToString(oldState) + " changing to FINISHED";
```
dstoc: need to do a pass converting to single quotes

test/test-runner.html
```diff
@@ -297,100 +315,154 @@ <h1 id=status>Pending</h1>
...
+
+    processResults(test);
+
     break;
-  }
 
-  function testSort(a, b) {
-    return a.id.localeCompare(b.id);
+  case States.POSTING:
+    if (!(oldState == States.FINISHED)) {
```
dstoc: use !=

test/bootstrap.js
mithro: Done.

test/bootstrap.js
mithro: Done.

test/bootstrap.js
mithro: Actually that comment is totally wrong.... Fixed.

test/bootstrap.js
mithro: Done.

test/bootstrap.js
mithro: Done.

test/bootstrap.js
mithro, line 595: Which one?

test/bootstrap.js
mithro: I find requestAnimationFrame non-intuitive. Unlike setTimeout/setInterval to keep getting called you have to raf() in each rafCallback, hence to stop you have to /do nothing/ which is pretty hard to document in the code.

Maybe the comment could be better?

test/test-runner.html
mithro: Done.

test/test-runner.html
mithro: Fixed indent.

As window.scroll took an array. Moved the code around a little so it's less confusing.

test/bootstrap.js
dstoc, line 595: The line this comment is on, 583

test/test-runner.html
mithro, line 281: Although we don't use closure compiler at the moment, it does some nice things like checking you have covered all the cases in a switch statement, optimizing the names down to smallest possible, etc.

test/test-runner.html
mithro: Fixed double to single quotes everywhere...

test/test-runner.html
mithro: Done.

test/test-runner.html
dstoc, line 281: @enum {string} ?

test/test-runner.html
mithro: Done.

test/test-runner.html
mithro: Done.

test/test-runner.html
mithro, line 281: Numbers should be faster in theory and seem to be preferred. Since we don't use the compiler it's probably moot anyway?

test/test-runner.html
mithro: I was using the 

```
if (!(<condition))
```

because it was easier to add more conditions as needed

test/test-runner.html
shans, line 281: We don't ever intend to use the compiler, no :)

Hence, I think Doug is right. We should use strings for readability. We're not switching in a tight loop or anything so I don't think performance is going to be a problem.

test/test-runner.html
shans: I don't understand this rationale. What is an example of a condition you might want to add later?

test/test-runner.html
mithro, line 281: Done.

test/test-runner.html
mithro: Was trying to be consistent with later in the file, the first statement in the switch statement is an inverted if statement. See later in the file:
- https://github.com/mithro/web-animations-js/blob/7d02b0f20e9652836cbb0b918d1055416c6f324d/test/test-runner.html#L369 
- https://github.com/mithro/web-animations-js/blob/7d02b0f20e9652836cbb0b918d1055416c6f324d/test/test-runner.html#L407

mithro:
PTAL.

test/test-runner.html
shans: Personally I'd change the first linked example to (a != FOO) && (a != BAR) && (a != FAR).

I don't understand what the second example does, which suggests that a refactor to introduce a test function is probably a good idea.

These are asides though; consistency between single-clause and multiple-clause boolean statements is not something that I would consider cogent to making style decisions.

mithro: Changed.

mithro:
PTAL.

PR #365, opened 2013-07-18
Fixes to web-animations.js for Firefox/Chrome differences.

mithro:


web-animations.js
```diff
@@ -881,7 +878,39 @@ var TimingEvent = function(token, target, type, localTime, timelineTime, iterati
   this.seeked = seeked ? true : false;
 }
 
-TimingEvent.prototype = Object.create(Event.prototype);
+TimingEvent.prototype = Object.create(Event.prototype, {
+  target: {
+    get: function() {
+      return this._target;
+    },
+    set: function(target) {
```
dstoc: Do we need setters for target/type?

dstoc:
lgtm

web-animations.js
mithro: I guess we should just set this._target in the constructor?

dstoc: I think you already did that. Remove these and you can merge.

mithro: Appears I did. Done!

PR #367, opened 2013-07-18
Removing some test complexity.

mithro:
- Only auto-test-\* load check files.
- Tests are disabled by renaming them to disable-*


test/bootstrap.js
```diff
@@ -65,8 +65,21 @@ function hasFlag(flag) {
   return thisScript && thisScript.getAttribute(flag) !== null;
 }
 
-function isUnitTest() {
-  return /unit-test[^\\\/]*\.html$/.exec(location.pathname);
+function testType() {
+  var p = location.pathname;
+  p.replace(/^disabled-/, '');
```
dstoc: Stale code? This doesn't appear to do anything.

test/bootstrap.js
```diff
@@ -65,8 +65,21 @@ function hasFlag(flag) {
...
+
+  if (/auto-test[^\\\/]*\.html$/.exec(p)) {
```
dstoc: [^\\/] should just be [^\/], escaping inside // is different

dstoc: OK, but a bit verbose... This could just be:

var match = /(auto|impl|manual|unit)-test[^\/]*$/.exec(p);
return match ? match[1] : 'unknown';

test/bootstrap.js
mithro: Fixed.

test/bootstrap.js
mithro: I think [^\\/] meant not '\' and not '/' ?

mithro: Converted to your regex method.

mithro: Fixed [^\\/]

dstoc:
> Actually, it is what the original author was doing (but he left no comments
> explaining why). I was just doing the same :)

Ahh. Sorry!

PR #370, opened 2013-07-23
Support CSS keyword values (aliases and non-transitional)

alancutter:
This patch is a replacement for the "support initial keyword" patch. Turns out it wasn't too big a step to generalise it to arbitrary keywords for animatable values.


web-animations.js
```diff
@@ -2899,6 +2891,8 @@ var colorType = {
   }
 };
 
+var colorType = addKeywordsToType(keywordlessColorType, ['currentColor']);
```
dstoc: since we don't need keyworlessColorType by itself, can we change this to

```
var colorType = typeWithKeywords(['currentColor'], {
  zero: ...,
  add: ...,
});
```

web-animations.js
```diff
@@ -3539,58 +3533,78 @@ var transformType = {
...
+  borderTopColor: colorType,
+  borderTopLeftRadius: percentLengthType,
+  borderTopRightRadius: percentLengthType,
+  borderTopWidth: lengthType,
+  bottom: percentLengthAutoType,
+  clip: addKeywordsToType(rectangleType, ['auto']),
+  color: colorType,
+  cx: lengthType,
+  fontSize: addKeywordsToType(percentLengthType, ['smaller', 'larger']), // TODO: Handle these properly.
+  fontWeight: addKeywordsToType(fontWeightType, ['lighter', 'bolder']), // TODO: Handle these properly.
```
dstoc: Didn't we handle them properly before?

dstoc: No, I was confused with normal/bold. Where did they go?

web-animations.js
```diff
@@ -3539,58 +3533,78 @@ var transformType = {
...
+  height: percentLengthAutoType,
+  left: percentLengthAutoType,
+  letterSpacing: addKeywordsToType(lengthType, ['normal']),
+  lineHeight: percentLengthType, // TODO: should be both number and percentLength
+  marginBottom: lengthAutoType,
+  marginLeft: lengthAutoType,
+  marginRight: lengthAutoType,
+  marginTop: lengthAutoType,
+  maxHeight: addKeywordsToType(percentLengthType,
+    ['none', 'max-content', 'min-content', 'fill-available', 'fit-content']),
```
dstoc: Indent at least 4 spaces when wrapping.

test/testcases/auto-test-initial.html
```diff
@@ -0,0 +1,208 @@
...
+word word word word</pre>
+
+
+<script src="../bootstrap.js"></script>
+<script>
+"use strict";
+
+
+document.timeline.play(new Animation(document.querySelector('#allToInitial'), [
+  {
```
dstoc: you can put { on the previous line and remove indentation

web-animations.js
```diff
@@ -2153,7 +2153,33 @@ var interpArray = function(from, to, f, type) {
   return result;
 };
 
-// TODO: This type does not handle 'inherit'.
+var addKeywordsToType = function(type, keywords) {
+  var isKeyword = {};
+  for (var i in keywords) {
+    isKeyword[keywords[i]] = true;
+  }
+  return createObject(type, {
```
dstoc, line 2161: Why createObject? Does type really have any other functions attached?

alancutter, line 2161: Just the zero() function though it's debatable whether that needs to be brought along or not.

web-animations.js
```diff
@@ -3647,9 +3755,15 @@ var fromCssValue = function(property, value) {
   if (value === cssNeutralValue) {
     return rawNeutralValue;
   }
+  if (value === 'inherit') {
+    return value;
+  }
   // Currently we'll hit this assert if input to the API is bad. To avoid this,
   // we should eliminate invalid values when normalizing the list of keyframes.
   // See the TODO in isSupportedPropertyValue().
+  if (property in propertyValueAliases) {
```
dstoc: This should probably go above the comment talking about the assert below.

web-animations.js
```diff
@@ -3600,6 +3618,87 @@ var svgProperties = {
...
+  backgroundPosition: { initial: '0% 0%' },
+  borderBottomColor: { initial: 'currentColor' },
+  borderBottomLeftRadius: { initial: '0px' },
+  borderBottomRightRadius: { initial: '0px' },
+  borderBottomWidth: borderWidthAliases,
+  borderLeftColor: { initial: 'currentColor' },
+  borderLeftWidth: borderWidthAliases,
+  borderRightColor: { initial: 'currentColor' },
+  borderRightWidth: borderWidthAliases,
+  borderSpacing: { initial: '2px' }, // Spec says this should be 0 but in practise it is 2px.
```
dstoc, line 3642: Does this depend on UA? Can we detect the correct value?

web-animations.js
```diff
@@ -3600,6 +3618,87 @@ var svgProperties = {
...
+  borderTopColor: { initial: 'currentColor' },
+  borderTopLeftRadius: { initial: '0px' },
+  borderTopRightRadius: { initial: '0px' },
+  borderTopWidth: borderWidthAliases,
+  bottom: { initial: 'auto' },
+  clip: { initial: 'rect(0px, 0px, 0px, 0px)' },
+  color: { initial: 'black' }, // Depends on user agent.
```
dstoc, line 3649: Can we detect it?

web-animations.js
```diff
@@ -3600,6 +3618,87 @@ var svgProperties = {
...
+    'x-small': '75%',
+    'small': '89%',
+    'medium': '100%',
+    'large': '120%',
+    'x-large': '150%',
+    'xx-large': '200%',
+  },
+  fontWeight: {
+    initial: '400',
+    normal: '400',
```
dstoc, line 3662: Ahh, I see.

web-animations.js
```diff
@@ -3647,9 +3755,15 @@ var fromCssValue = function(property, value) {
...
     return rawNeutralValue;
   }
+  if (value === 'inherit') {
+    return value;
+  }
   // Currently we'll hit this assert if input to the API is bad. To avoid this,
   // we should eliminate invalid values when normalizing the list of keyframes.
   // See the TODO in isSupportedPropertyValue().
+  if (property in propertyValueAliases) {
+    value = propertyValueAliases[property][value] || value;
```
dstoc: I think you need to check whether the result is undefined rather than using ||. eg. for propertyValueAliases['transform']['none'] which is falsy

web-animations.js
alancutter, line 3649: Without adding elements to the body and getting their computed style it's a bit tricky.

dstoc, line 3649: That should be fine. File a bug and we can fix later if you like.

web-animations.js
alancutter: Good catch!

web-animations.js
alancutter, line 3649: Will file bug. Implement dodgy solution?

dstoc, line 3649: I just meant leave it as is. We can detect it somehow later.

alancutter, line 3649: Okay, filed bug: #371.

web-animations.js
alancutter, line 3662: Good. (:

web-animations.js
alancutter, line 3642: Filed bug: #371.

web-animations.js

test/testcases/auto-test-initial.html
alancutter: Done.

web-animations.js

web-animations.js

PR #374, opened 2013-07-24
Support animating shorthand properties

alancutter:


web-animations.js
```diff
@@ -1585,6 +1585,107 @@ PathAnimationEffect.prototype = createObject(AnimationEffect.prototype, {
   }
 });
 
+var shorthandToLonghand = {
```
dstoc, line 1588: Can we determine these automatically?

I think there are only a few special cases where foo is not a shorthand of foo-bar.

dstoc, line 1588: See: http://lists.w3.org/Archives/Public/www-style/2013Jul/0329.html

alancutter, line 1588: It's certainly possible but we'd significantly lose code clarity.

dstoc, line 1588: We're going to need to support all shorthands at some point, I don't want to have to maintain a massive list.

Shall we file a bug to support all properties by generating this list?

alancutter, line 1588: We only need to expand the shorthands that alias animatable longhand properties rather than list all shorthands, the rest are already handled.
Filed #376.

PR #380, opened 2013-07-26
Disable assert and remove use of hasOwnProperty -- they are slow

dstoc:


web-animations.js
```diff
@@ -2063,7 +2065,7 @@ KeyframeAnimationEffect.prototype = createObject(AnimationEffect.prototype, {
       }
       this._cachedProperties = [];
       for (var p in properties) {
-        if (properties.hasOwnProperty(p)) {
+        if (p in properties) {
```
alancutter: Do we still need this check?

dstoc: Nope! Refactored.

web-animations.js
```diff
@@ -2096,22 +2096,22 @@ KeyframeInternal.prototype = {
     return this.rawValues[property];
   },
   addPropertyValuePair: function(property, value) {
-    console.assert(!this.cssValues.hasOwnProperty(property));
+    ASSERT_ENABLED && console.assert(!this.cssValues.hasOwnProperty(property));
     this.cssValues[property] = value;
   },
   hasValueForProperty: function(property) {
-    return this.cssValues.hasOwnProperty(property);
+    return property in this.cssValues;
```
alancutter, line 2103: I would add an assert here that hasOwnProperty() matches 'in' to be safe.

dstoc, line 2103: It does, since we initialize (line 2086) and populate (line 2100) cssValues ourselves

alancutter, line 2103: I see, all good then.

PR #386, opened 2013-08-01
Fixing the number matching logic in assert_styles (and associated tests).

mithro:
The following things were broken with the number matching logic;
- Scientific notation did not work (1.0e-9) was extracted as 1.0
- When the actual was shorter than expected you'd get a weird message about values not being NaN 
- When the actual was longer than expected, it would be doing an accidental startswith comparison.

This was discovered when trying to add fix precision checking to assert_styles.


test/bootstrap.js
```diff
@@ -167,6 +167,25 @@ function _assert_style_get(style, i) {
   }
 }
 
+
+/**
+ * Extract all the numeric values from a string.
+ */
+function _extract_numbers(v) {
```
dstoc: shouldn't need the split / reverse. you can do multi-match with something like this:

```
var re = /.../g;
var match;
var result = [];
while (match = re.exec(input)) {
  result.push(Number(match[0]));
}
return result;
```

mithro: I was originally trying to do something with the non-numeric parts to but that turned out to be harder then it looked.

Will change to your method.

test/testcases/auto-test-text-shadow-checks.js
```diff
@@ -12,9 +12,9 @@ timing_test(function() {
     assert_styles('body', {'textShadow':'rgba(64, 0, 0, 0.247059) 25px 25px 0px, rgba(64, 0, 0, 0.247059) 0px 25px 2px'});
   }, "Autogenerated");
   at(2, function() {
-    assert_styles('body', {'textShadow':'rgb(0, 0, 0) 0px 0px 0px'});
+    assert_styles('body', {'textShadow':'rgba(0, 0, 0) 0px 0px 0px, rgba(0, 0, 0) 0px 0px 0px'});
```
dstoc, line 15: If we merge https://github.com/web-animations/web-animations-js/pull/388 first, we probably don't need this.

mithro, line 15: Happy to rebase if we merge #388 first.

dstoc:
lgtm

PR #393, opened 2013-08-06
Add test failure expectations for Firefox and make Firefox gating.

mithro:


test/testcases/test-update-state.html
```diff
@@ -14,6 +14,17 @@
...
+
+<script>
+var expected_failures = {
+  'Updated style should be visible despite unrelated errors at t=5s': {
+    firefox: true,
+    message: "Floating point issues."
+  }
+};
+</script>
+Updated style should be visible
```
dstoc, line 26: Is this meant to be here?

dstoc:
lgtm

test/testcases/test-update-state.html
mithro, line 26: Fixed the description;

Fail    Updated style should be visible despite unrelated errors at t=5s    assert_equals: bottom is not 0px, actually 500px expected 0 but got 500

PR #398, opened 2013-08-07
Bunch of small changes to the generator to make it nicer to use.

mithro:


test/test-generator.html
```diff
@@ -207,14 +212,16 @@
...
     }
-    printCheck(checks[x].initSelctor, i, propslist, time);
+    outputText += printCheck(checks[x].initSelctor, i, propslist, time);
   }
+  outputText += '  });';
+  var output = document.getElementById("checks");
+  output.innerHTML += outputText + "<br>";
 }
 
 function printCheck(object, number, properties, time){
```
dstoc, line 222: it doesn't really 'print', perhaps formatCheck ?

dstoc:
lgtm

test/test-generator.html
mithro, line 222: I didn't name it :), probably should fix all the names in this code sometime....

PR #399, opened 2013-08-07
Making the log elements <pre> so they look nicer.

mithro:


dstoc:
You should be able to do this in CSS without creating additional elements.
eg. white-space: pre; font-family: monospace;

mithro:
Fixed. PTAL.

PR #401, opened 2013-08-08
Fixing a number of problems which where breaking tests under Firefox.

mithro:


test/testcases/disabled-test-dynamic-load.html
```diff
@@ -14,7 +14,14 @@
 limitations under the License.
 -->
 
-<!DOCTYPE html>
+<!--
+This test is disabled because in Firefox the page is consided loaded before
+web-animations.js is loaded and hence the test will fail as web-animations.js
```
dstoc: I don't understand this comment, the test waits for the load event before it loads web-animations.js

mithro: Moved to different pull-request.

PR #416, opened 2013-08-14
Attempt at introducing basic player cleanup.

shans:


steveblock:
LGTM

dstoc:
I think we need a test like in test-update-state.html where we have an animation whose Player has been removed from the list, but then we modify the timing in such a way that it would then be in effect (and the effect should be visible immediately).

web-animations.js
```diff
@@ -707,6 +721,9 @@ TimedItem.prototype = {
     return this.parent === null ? effectivePlaybackRate :
         effectivePlaybackRate * this.parent._netEffectivePlaybackRate();
   },
+  _hasEffect: function() {
+    return this._isCurrent() || this.specified.fillMode !== 'none';
```
dstoc, line 731: Can we just implement this as defined in the spec?

inEffect = activeTime != null;

We want to keep players in this list for anything which is current, or in effect.

shans, line 731: That won't work as soon as the new fill mode semantics are implemented.

shans, line 731: Hmm, actually, maybe it will.

shans, line 731: I tried this. It doesn't work because we're not interested in whether the player has effect right _now_, we're interested in whether it could have an effect in the _future_. Accordingly, I've renamed _hasEffect to _hasFutureEffect to make this clearer.

shans:
One of the tests in test-player-update.html already tests that inactive players were made active again by changing the player's current time. I've beefed it up a bit to test that the player is active immediately after the change, and that the styles are applied immediately.

I also added a similar test that modifies the source TimedItem's specified timing instead. This uncovered a bug, which I've fixed :)

web-animations.js
dstoc, line 731: hasFutureEffect = isCurrent || isInEffect 
and isInEffect = activeTime != null

But maybe that still depends on the new fill mode behavior being implemented.

PR #418, opened 2013-08-15
Two small fixes so run-tests.py works on OS X.

mithro:
- platform.system() returns "Darwin" not "mac".
- Don't hardcode the google-chrome binary to /usr/bin


mikelawther:
lgtm

PR #419, opened 2013-08-15
Fix bug causing players to not play when started in an onend event 

alancutter:
Fix bug causing players to not play when started in an onend event and all other players have finished.


dstoc:
@shans -- was it intentional that events are fired synchronously?

dstoc:
@shans said no.

So let's fix this by making events fire asynchronously.

alancutter:
Happy to make events fire asynchronously, this will need to wait for #420 to land as the setTimeout calls need to wrap the callback calls as tight as possible to work correctly.

PR #420, opened 2013-08-16
Implement addEventListener and removeEventListener interface on TimedItem

alancutter:


web-animations.js
```diff
@@ -708,35 +709,64 @@ TimedItem.prototype = {
...
+  },
+  addEventListener: function(type, fun) {
+    if (typeof fun !== 'function' || !(type !== 'start' || type !== 'iteration' ||
+      type !== 'end' || type !== 'cancel')) {
+      return;
+    }
+    if (!isDefinedAndNotNull(this._handlers[type])) {
+      this._handlers[type] = [];
+    }
+    this._handlers[type].push(fun);
```
dstoc: What if 'fun' is already in the list, should it be added again?

web-animations.js
```diff
@@ -708,35 +709,64 @@ TimedItem.prototype = {
...
   get oncancel() {
-    return this._cancelHander;
+    return this._getSingleHandler('cancel');
+  },
+  _setSingleHandler: function(type, fun) {
+    this._handlers[type] = (typeof fun === 'function') ? [fun] : null;
+    this._newHandler(fun);
+  },
+  _getSingleHandler: function(type) {
+    return this._handlers[type] ? this._handlers[type][0] : null;
```
dstoc: Is this how regular dom works? (returning the first added by addEventListener)

web-animations.js
alancutter: It should not, changing code...

web-animations.js
alancutter: This is incorrect behaviour, changing implementation to match behaviour shown in Blink and Firefox.

web-animations.js
```diff
@@ -730,50 +732,114 @@ TimedItem.prototype = {
...
-      isDefinedAndNotNull(this._cancelHandler);
+  _getOnHandler: function(type) {
+    if (isDefinedAndNotNull(this._onHandler[type])) {
+      return this._onHandler[type].func;
+    }
+    return null;
+  },
+  addEventListener: function(type, func) {
+    if (typeof func !== 'function' || !(type !== 'start' || type !== 'iteration' ||
+      type !== 'end' || type !== 'cancel')) {
```
dstoc: Indentation

web-animations.js
```diff
@@ -730,50 +732,114 @@ TimedItem.prototype = {
...
+      return;
+    }
+    if (!isDefinedAndNotNull(this._handlers[type])) {
+      this._handlers[type] = [func];
```
dstoc: Why not initialize _handlers[type] to [] and fall through to the logic below?

web-animations.js
```diff
@@ -730,50 +732,114 @@ TimedItem.prototype = {
...
+  },
+  _callHandlers: function(type, eventFunction) {
+    var onIndex =  -1;
+    if (isDefinedAndNotNull(this._onHandlers[type])) {
+      onIndex =  this._onHandlers[type].index;
+    }
+    var handlersLength = (this._handlers[type] || []).length;
+    for (var i = 0; i < handlersLength; i++) {
+      if (onIndex === i) {
+        this._onHandlers[type].callback(eventFunction());
```
dstoc: We probably only want to create the event once and pass it to all the callers.

web-animations.js
```diff
@@ -730,50 +732,114 @@ TimedItem.prototype = {
...
+      }
+      this._handlers[type][i].call(this, eventFunction());
+    }
+    if (onIndex >= handlersLength) {
+      this._onHandlers[type].callback(eventFunction());
```
dstoc: callback.call(this, ... ) ?

web-animations.js
```diff
@@ -819,16 +885,20 @@ TimedItem.prototype = {
     }
     var startTime = this.startTime + this.specified.startDelay;
 
-    if (isDefinedAndNotNull(this.onstart)) {
+    if (this._hasHandlersForEvent('start')) {
       // Did we pass the start of this animation in the forward direction?
       if (fromTime <= startTime && toTime > startTime) {
-        this.onstart(new TimingEvent(constructorToken, this, 'start',
-            this.specified.startDelay, toGlobal(startTime), firstIteration));
+        this._callHandlers('start', (function() {
```
dstoc: Why are we passing this closure instead of the event? We've already checked that there are handlers for the event.

web-animations.js
```diff
@@ -730,50 +732,114 @@ TimedItem.prototype = {
...
+      return this._hasHandlersForEvent(type);
+    }).bind(this));
+  },
+  _hasHandlersForEvent: function(type) {
+    return (isDefinedAndNotNull(this._handlers[type]) &&
+        this._handlers[type].length > 0) ||
+        isDefinedAndNotNull(this._onHandlers[type]);
+  },
+  _callHandlers: function(type, eventFunction) {
+    var onIndex =  -1;
```
dstoc: extra space here

web-animations.js
```diff
@@ -730,50 +732,114 @@ TimedItem.prototype = {
...
+    if (isDefinedAndNotNull(this._onHandlers[type])) {
+      onIndex =  this._onHandlers[type].index;
```
dstoc: extra space here

alancutter: Fixed.

web-animations.js
alancutter: Fixed.

web-animations.js
alancutter: This is to allow the callback caller function generate a new event object for each callback call (preventing callbacks from modifying the event object received by other callbacks).

web-animations.js
alancutter: onevents execute using the global context for this while addEventListeners get the target element.
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget.addEventListener#The_value_of_this_within_the_handler

web-animations.js
alancutter: I decided against doing so as a callback may modify the event object, do you think this issue should be overlooked?

web-animations.js

web-animations.js
alancutter: Fixed.

web-animations.js
alancutter: Ah, it seems I'm mistaken in regard to this behaviour. The event object should be shared by all event handlers it appears.
Fixed.

dstoc:
Did you mean to use a different account?

web-animations.js
dstoc: I'm not sure. That's not the behavior I saw in Chrome when I was experimenting.

In the example on MDN:

```
onclick="modifyText();"
```

That is setting the onclick _attribute_ to the _string_ "modifyText()". Not to the function modifyText.

That would be equivalent to the following:

```
foo.setAttribute("onclick", "modifyText()");
```

But that doesn't apply to TimedItems as they are not Nodes.

web-animations.js
```diff
@@ -730,50 +732,109 @@ TimedItem.prototype = {
...
+      this._onHandlers[type].index -= 1;
+    }
+    if (this.player) {
+      this.player._checkForHandlers();
+    }
+  },
+  _hasHandlers: function() {
+    return ['start', 'iteration', 'end', 'cancel'].some((function(type) {
+      return this._hasHandlersForEvent(type);
+    }).bind(this));
```
dstoc: I fear that JS engines are unlikely to be able to optimize this.

Seems safer to just write this as return this._hasHandlersForEvent('start') || ...

web-animations.js
```diff
@@ -730,50 +732,109 @@ TimedItem.prototype = {
...
+        this._handlers[type].length > 0) ||
+        isDefinedAndNotNull(this._onHandlers[type]);
+  },
+  _callHandlers: function(type, event) {
+    var onIndex = -1;
+    if (isDefinedAndNotNull(this._onHandlers[type])) {
+      onIndex = this._onHandlers[type].index;
+    }
+    var handlersLength = (this._handlers[type] || []).length;
+    for (var i = 0; i < handlersLength; i++) {
```
dstoc: how about:

```
for (var i = 0; i <= handlersLength; i++) {
  if (onIndex == i) {
    this._onHandlers[type]...
  }
  if (i < handlersLength) {
    this._handlers[type][i]...
  }
```

web-animations.js

web-animations.js
alancutter: Honestly I'm not sure much is being gained by removing a function call duplication at the cost of adding an additional check inside the for loop.

alancutter:
Whoops, I was committing with my personal account. Updated local repo config.

web-animations.js
dstoc: I made this suggestion because I found the existing code hard to follow. The check is negligible here. We should prefer code which is easy to read and verify.

web-animations.js
dstoc: I think the comment above got lost.

alancutter: The code should be updated now.

dstoc: I still see:

```
...
  this._onHandlers[type].callback(event);
}
if (i < handlersLength) {
  this._handlers[type][i].call(this, event);
```

alancutter:
Rebased change onto timing input renames.

dstoc:
I still don't see what happened with https://github.com/web-animations/web-animations-js/pull/420#discussion-diff-5811666

web-animations.js
alancutter: Aaah, confusing! I saw this in an email but not in the outdated diff.
I see your comment regarding the this value now.
Updated accordingly.

PR #425, opened 2013-08-19
Make event handling asynchronous

alancutter:
When players are played in the end event of the last player they do not get played.
Making the events fire asynchronously fixes the problem by allowing player registration to occur at the appropriate time.

### DIFF
--- a/test/testcases/manual-test-add-player-onend.html
+++ b/test/testcases/manual-test-add-player-onend.html
@@ -45,8 +45,6 @@ var element = document.querySelector('#target');
 var animation = new Animation(element, {left: '50px'}, 0.5);
 animation.addEventListener('end', function() {
   var nextAnimation = new Animation(element, {left: '100px'}, 0.5);
-  nextAnimation.addEventListener('start', function() {
-  });
   nextAnimation.addEventListener('end', function() {
     test.step(function() { assert_styles("#target", {left:'100px'}); });
     test.done();
--- a/test/testcases/manual-test-add-player-onend.html
+++ b/test/testcases/manual-test-add-player-onend.html
@@ -42,16 +42,13 @@ limitations under the License.
 
 var test = async_test('Check second animation has fired after being triggered inside onend callback');
 var element = document.querySelector('#target');
-var animation = new Animation(element, { left: '50px' }, 0.5);
+var animation = new Animation(element, {left: '50px'}, 0.5);
 animation.addEventListener('end', function() {
-  console.log('testpants');
-  var nextAnimation = new Animation(element, { left: '100px' }, 0.5);
+  var nextAnimation = new Animation(element, {left: '100px'}, 0.5);
   nextAnimation.addEventListener('start', function() {
-    console.log('testDuck');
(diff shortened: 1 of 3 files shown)


alancutter:
The testharness can't handle two levels of asynchronousicity unfortunately so the test for this bug has to be manual.

test/testcases/manual-test-add-player-onend.html
```diff
@@ -0,0 +1,62 @@
...
+
+<script src="../bootstrap.js"></script>
+<script>
+'use strict';
+
+var test = async_test('Check second animation has fired after being triggered inside onend callback');
+var element = document.querySelector('#target');
+var animation = new Animation(element, { left: '50px' }, 0.5);
+animation.addEventListener('end', function() {
+  console.log('testpants');
```
dstoc: Should this be 'testPants' for consistency?

dstoc: On second thoughts best not to have logging at all ;)

test/testcases/manual-test-add-player-onend.html
```diff
@@ -0,0 +1,62 @@
...
+<div id="expected"></div>
+<div id="target"></div>
+
+<script src="../bootstrap.js"></script>
+<script>
+'use strict';
+
+var test = async_test('Check second animation has fired after being triggered inside onend callback');
+var element = document.querySelector('#target');
+var animation = new Animation(element, { left: '50px' }, 0.5);
```
dstoc: don't need spaces in a single line object initializer:

```
{left: '50px'}
```

test/testcases/manual-test-add-player-onend.html
```diff
@@ -0,0 +1,62 @@
...
+  nextAnimation.addEventListener('end', function() {
+    console.log('testChicken');
+    test.step(function() { assert_styles("#target", {'left':'100px'}); });
+    test.done();
+  });
+  document.timeline.play(nextAnimation);
+});
+document.timeline.play(animation);
+
+timing_test(function() {at(1, function() {});}, 'Force testharness to execute to 1 second.');
```
dstoc, line 56: Hmm, does this still not work when named unit-test-*.html ? I had assumed all of the timing test logic would be disable then.

alancutter, line 56: Tried as unit-test-*, still failed to execute second end event handler and timed out the test.

test/testcases/manual-test-add-player-onend.html
alancutter: Done.

test/testcases/manual-test-add-player-onend.html
alancutter: The non-camel cased debug print comes back to haunt me!
Yes it should be testPants and yes it should not be there. :)

test/testcases/manual-test-add-player-onend.html
alancutter, line 56: Filed #426 for this issue.

PR #429, opened 2013-08-20
Fixing extra comma.

mithro:


dstoc:
Was there a reason to make this change? Style or compat problem? I notice there's the same on line 686.

Even trivial changes like this should still be reviewed.

mithro:
IE dies with an error on this line without it.

Don't know why it doesn't barf on line 686?

PR #430, opened 2013-08-20
Cause errors to be raise outside the TimingTest loop.

mithro:
This means even if there is an error, the rest of the tests / callbacks are run.

### DIFF
--- a/test/bootstrap.js
+++ b/test/bootstrap.js
@@ -558,7 +558,14 @@ TestTimelineGroup.prototype.call = function() {
     var status_ = statuses[statuses.length - callbacks.length - 1];
 
     if (typeof callback == 'function') {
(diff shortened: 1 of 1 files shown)


dstoc:
Why do we want to do this?

mithro:
See the comment. It means that an error doesn't stop the rest of the tests
from executing. Just a try/catch would then how the errors.

dstoc:
> See the comment. It means that an error doesn't stop the rest of the tests
> from executing. Just a try/catch would then how the errors.

I mean, why wouldn't we want to stop the rest of the tests/callbacks from executing?

mithro:
When viewed in the test-runner the whole page displays a timeout rather then providing a useful error message.

PR #435, opened 2013-08-21
Another attempt at making the state transitions more reliable

mithro:


test/test-runner.html
```diff
@@ -505,8 +514,11 @@ <h1 id=status>Pending</h1>
 
     document.querySelector('#results').appendChild(test);
     test.querySelector('iframe').contentWindow.onload = function() {
-      changeTestState(this, 'LOADED');
-      runTestsIfLoaded();
+      // You can get multiple onload events, only deal with the first one.
```
dstoc, line 517: This is because below we now set the src when the test times out?

PR #438, opened 2013-08-21
Make the update of test status async.

mithro:
If some problem happens in updating state, the test status will still update (where as previously when an exception occurred the state would never update leaving you blind).


dstoc:
The change seems fine, but the reason to make this change is missing from the description.

mithro:
Description updated....

PR #440, opened 2013-08-21
Rolio demo

shans:


dstoc:
lgtm, but you may want to clean up for JS style:
- prefer single quotes
- spaces around binary operators

PR #445, opened 2013-08-22
Implement chained timing functions

steveblock:


birtles:
Wow, that was fast :)

One thing I just noticed though is that from a quick skim of the test cases this seems to allow specifying easingTimes as a space-separated string but currently the API only lets you specify "align" "distribute" or a sequence of numbers.

steveblock:
> Wow, that was fast :)

It was just good timing - I'd just started when you sent out the spec update!

> currently the API only lets you specify "align" "distribute" or a sequence of numbers.

Ah, I hadn't even noticed that. What's the reason for the restriction? All the required logic is already there.

mithro:
Test currently fail....

test/testcases/unit-test-chained-timing-function.html
```diff
@@ -0,0 +1,239 @@
+<!--
```
dstoc: This should be impl-test since it uses _WebAnimationsTestingUtilities

birtles:
> > currently the API only lets you specify "align" "distribute" or a sequence of numbers.
> 
> Ah, I hadn't even noticed that. What's the reason for the restriction? All the required logic is already there.

The restriction is that you can't specify a sequence of numbers as a string. The reason is that working with numbers is easier than having to parse the string and put it back together again. And we don't support both because it would complicate the code for the author when they go to read back values, e.g.: test if string or sequence, if string test if keyword value or space-separated list of numbers, if sequence of numbers parse etc. etc.

test/testcases/test-chained-timing-function.html
```diff
@@ -0,0 +1,77 @@
+<!--
```
dstoc, line 1: Should also add a few test cases to auto-test-timing-functions.html

web-animations.js
```diff
@@ -2248,7 +2242,122 @@ TimingFunction.createFromString = function(spec, timedItem) {
...
+    for (var i = 0; i <= timingFunctions.length; i++) {
+      points.push(i / timingFunctions.length);
+    }
+  }
+  ASSERT_ENABLED && console.assert(Array.isArray(points) && points.length > 0);
+
+  var result = [];
+  var lastTimingFunction;
+  for (var i = 1; i < points.length; i++) {
+    var range = { min: points[i - 1], max: points[i] };
```
dstoc: no spaces inside single line object initializer {foo: bar}

dstoc:
lgtm

steveblock:
Updated to fix style, add more tests and fix a couple of bugs that were breaking existing tests.

steveblock:
> The restriction is that you can't specify a sequence of numbers as a string.

Got it: I misunderstood your first comment. I've updated the logic to require the list of points to be an array of numbers.

Not yet for landing, as Shane has also started an implementation and we're going to pick the best bits of both.

steveblock:
I've talked to Shane and I'm going to merge aspects of our implementations. In the mean time, we can land the preparatory commits, which I've moved to https://github.com/web-animations/web-animations-js/pull/452

steveblock:
Updated to use the best bits of Shane's and my implementations, fixed a few more bugs (from both) and added more tests.

web-animations.js
```diff
@@ -2305,39 +2326,200 @@ var TimingFunction = function() {
...
+
+  var normalizedChain = [];
+  var lastTimingFunction = chain[chain.length - 1];
+  for (var i = 1; i < normalizedPositionList.length; i++) {
+    var range = {
+      min: normalizedPositionList[i - 1],
+      max: normalizedPositionList[i]
+    };
+    var timingFunction = i - 1 < chain.length ?
+        chain[i - 1] : lastTimingFunction.clone();
```
dstoc, line 2363: Why do we need to use clone() here?

dstoc, line 2363: Oh, because PacedTimingFunction is based on the range.

web-animations.js
```diff
@@ -2305,39 +2326,200 @@ var TimingFunction = function() {
...
+  }
+  return normalizedChain;
+};
+
+TimingFunction.createChain = function(easing, timedItem) {
+  var chain = [];
+  while (typeof easing === 'string') {
+    easing = easing.trim();
+    var result = TimingFunction.createComponentFromString(easing, timedItem);
+    if (result === null) {
```
dstoc: Does this happen when we can't parse a timing function? Shouldn't we fall back to linear in this case?

dstoc:
lgtm

web-animations.js
steveblock, line 2363: Right, see the last test in impl-test-chained-timing-function.html

web-animations.js
steveblock: Good catch Fixed.

PR #450, opened 2013-08-22
Need getter for TimingEvent.prototype.returnValue

morethanreal:


dstoc:
Fixed in https://github.com/web-animations/web-animations-js/pull/451

PR #452, opened 2013-08-23
Preparatory work for chained timing functions

steveblock:


shans:
LGTM.

dstoc:
Tests fail in firefox: http://postimg.org/image/u6oybw5iv/

steveblock:
Rebased and added failing expectations for Firefox (it looks like Firefox has fundamental problems with path effects)

steveblock:
I had to drop the commit to update the expected results for auto-test-timing-functions.html because I ran into rounding errors on Firefox and IE.

PR #454, opened 2013-08-28
Allow for editing of target element inline styles during animations

alancutter:


alancutter:
I think Travis can't handle the Unicode character in the branch name. :(

PR #455, opened 2013-08-28
Allow editing of inline styles during an animation

alancutter:


web-animations.js
```diff
@@ -4703,10 +4703,122 @@ CompositedPropertyMap.prototype = {
...
+    this._surrogateElement.style[property] = this._style[property];
+  }
+  // Delegate all property accessors to the surrogate element's inline style.
+  for (var property in this._style) {
+    if (cssStyleDeclarationAttribute[property] ||
+        cssStyleDeclarationMethod[property]) {
+      continue;
+    }
+    (function(animatedStyle, property) {
+      Object.defineProperty(animatedStyle, property, configureDescriptor({
```
dstoc: Do we really need to do this on every instance of AnimatedCSSStyleDeclaration? Perhaps we can just do it once on the prototype.

web-animations.js
```diff
@@ -4706,6 +4706,120 @@ CompositedPropertyMap.prototype = {
...
+      'Element must not already have an animated style attached.');
+  this._element = element;
+  this._surrogateElement = document.createElement('div');
+  this._style = element.style;
+  this._length = 0;
+
+  // Populate the surrogate element's inline style.
+  for (var i = 0; i < this._style.length; i++) {
+    var property = this._style[i];
+    this._surrogateElement.style[property] = this._style[property];
```
dstoc, line 4744: Do we need to copy every property, can't we do it on demand for the properties that are being animated?

web-animations.js
```diff
@@ -4706,6 +4706,120 @@ CompositedPropertyMap.prototype = {
...
+
+
+
+/** @constructor */
+var AnimatedCSSStyleDeclaration = function(element) {
+  ASSERT_ENABLED && assert(
+      !(element.style instanceof AnimatedCSSStyleDeclaration),
+      'Element must not already have an animated style attached.');
+  this._element = element;
+  this._surrogateElement = document.createElement('div');
```
dstoc, line 4736: It's not clear what purpose the surrogateElement serves. Add a comment or mention it in the description.

web-animations.js
alancutter: Quite right, done.

web-animations.js
alancutter, line 4744: Doing so would force the inline style accessor functions to be hand coded rather than automatically delegate and increase the overall complexity of the class.
This operation should be cheap and not cause layouts as is.

web-animations.js
alancutter, line 4736: Added comment.

web-animations.js
```diff
@@ -4706,6 +4706,124 @@ CompositedPropertyMap.prototype = {
...
+  }
+  (function(property) {
+    Object.defineProperty(AnimatedCSSStyleDeclaration.prototype, property,
+        configureDescriptor({
+          get: function() {
+            return this._surrogateElement.style[property];
+          },
+          set: function(value) {
+            this._surrogateElement.style[property] = value;
+            this._style[property] = value;
```
dstoc: This shouldn't happen if the property is currently animated.

alancutter: Done, added FIXME to indicate slightly incorrect behaviour between animation frames.

alancutter:
Latest Travis run is happy.

PR #459, opened 2013-08-28
Rename beforeValue and afterValue

alancutter:
This is to avoid a flaky V8 bug in beta/unstable Chrome on Travis.
Somehow beforeValue was being set as undefined: http://postimg.org/image/6ej865plh/


alancutter:
Looks like this patch failed, still getting the bug with the rename.
Will experiment further.

PR #460, opened 2013-08-28
Added changing parameter to src in script loader

alancutter:
This should prevent scripts from being cached by the browser.

### DIFF
--- a/test/bootstrap.js
+++ b/test/bootstrap.js
@@ -70,7 +70,6 @@ function getSync(src) {
 
 function loadScript(src, options) {
   // Add changing parameter to prevent script caching.
-  src += '?' + window.Date.now();
   options = options || {coverage: true};
   if (window.__resources__[src]) {
     document.write('<script type="text/javascript">eval(window.__resources__["'+src+'"]);</script>');
@@ -78,6 +77,7 @@ function loadScript(src, options) {
     instrument(src);
     loadScript(src);
   } else {
+    src += '?' + window.Date.now();
     document.write('<script type="text/javascript" src="'+ src + '"></script>');
   }
 }
--- a/test/bootstrap.js
+++ b/test/bootstrap.js
@@ -69,6 +69,8 @@ function getSync(src) {
 }
 
 function loadScript(src, options) {
+  // Add changing parameter to prevent script caching.
+  src += '?' + window.Date.now();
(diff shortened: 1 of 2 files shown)


test/bootstrap.js
```diff
@@ -69,6 +69,8 @@ function getSync(src) {
 }
 
 function loadScript(src, options) {
+  // Add changing parameter to prevent script caching.
+  src += '?' + window.Date.now();
```
dstoc: I think this breaks the caching that's done for code coverage, so that we don't have to load and instrument web-animations.js for every test

See line 75

dstoc: Maybe just disable it if it's cached in **resources** ?

mithro:
This also doesn't cover anything else which can change, IE the actual html page or bootstrap.js

test/bootstrap.js
alancutter: Done.

alancutter:
Added changing parameter to test file loading.
It's true that some things aren't covered by this but web-animations.js and the test files are covered which is where the majority of the changes in behaviour are going to occur.

test/bootstrap.js
```diff
@@ -69,6 +69,7 @@ function getSync(src) {
 }
 
 function loadScript(src, options) {
+  // Add changing parameter to prevent script caching.
```
dstoc, line 72: Move the comment to line 80

test/bootstrap.js
```diff
@@ -76,6 +77,7 @@ function loadScript(src, options) {
     instrument(src);
     loadScript(src);
   } else {
+    src += '?' + window.Date.now();
```
dstoc, line 80: window. is unnecessary and inconsistent with line 194

dstoc:
lgtm

test/bootstrap.js
alancutter, line 72: Done.

test/bootstrap.js
alancutter, line 80: Line 194 appears to be a blank block comment line.
There are two other instances of "window.Date" in this file.

dstoc, line 80: GitHub!! Line 194 was in another file, test-runner.html

OK to leave it as is given the other uses. But we should clean them all up at some point.

PR #468, opened 2013-08-30
Add try catch around flaky Travis bug crash site

alancutter:
Still at a complete loss as to how Travis is flakily crashing here, adding extra info to the crash information should help work this out.

Crash screenshot: http://postimg.org/image/y00j927cd/

### DIFF
--- a/web-animations.js
+++ b/web-animations.js
@@ -4611,8 +4611,13 @@ BlendedCompositableValue.prototype = createObject(
             this.fraction);
       },
       dependsOnUnderlyingValue: function() {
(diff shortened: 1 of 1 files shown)


dstoc:
Can you just keep submitting a retry on travis for this request until it fails?

alancutter:
Luck has not been on my side. I've only had one flaky failure on a commit made to debug it, all the other failures have been evasive and only occurring on other branches (quite regularly).
This try catch will be removed when the problem is solved.

dstoc:
OK, let's see what happens, again!

PR #469, opened 2013-08-30
Attach lazy retick to getComputedStyle after inline style change

alancutter:
Setting the inline style of an animated element causes the animation frame to completely re-evaluate so that getComputedStyle() can be accurate with respect to the new inline style and any animations ontop of it.

Users that set hundreds of inline styles per frame without calling getComputedStyle are punished with a useless animation tick after each set. This change fixes that.

### DIFF
--- a/web-animations.js
+++ b/web-animations.js
@@ -4779,7 +4779,7 @@ AnimatedCSSStyleDeclaration.prototype = {
             this._surrogateElement.style.getPropertyValue(property));
       }
     }
-    repeatLastTick();
+    this._inlineStyleChanged();
   },
   get length() {
     return this._surrogateElement.style.length;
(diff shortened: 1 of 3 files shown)


web-animations.js
```diff
@@ -4819,6 +4819,16 @@ AnimatedCSSStyleDeclaration.prototype = {
   _setAnimatedProperty: function(property, value) {
     this._style[property] = value;
     this._isAnimatedProperty[property] = true;
+  },
+  _inlineStyleChanged: function() {
+    maybeRestartAnimation();
```
dstoc, line 4824: Why is this needed? Perhaps it should have a return value to indicate whether a tick will happen or not?

alancutter, line 4824: Without calling maybeRestartAnimation() the element won't have its animation reapplied on the next frame if no animations are running and getComputedStyle() is never called.
I'm not sure what a return value would be used for.

dstoc, line 4824: I looked further, maybeRestartAnimation doesn't do what I thought it would do. I assumed it would only schedule at tick if one was required. So this seems OK.

alancutter, line 4824: Perhaps it should be named ensureTickScheduled(). :)

PR #497, opened 2013-11-07
Add inline style hack fallback (fixes Safari)

### LINKED ISSUE #492: Attempting to override getter element style fails in Safari

alancutter:
This change provides an alternate patch for animated inline styles.
If replacing element.style with an AnimatedCSSStyleDeclaration is not allowed (which is the case for Safari) it will instead patch the existing inline style object methods to behave in a similar way.
Property getters and setters are not patched and will not function normally while under animation with this fallback.

Fixes #492

### DIFF
--- a/web-animations.js
+++ b/web-animations.js
@@ -4769,20 +4769,24 @@ var retickThenGetComputedStyle = function() {
 };
 
 var originalGetComputedStyle = window.getComputedStyle;
+// This redundant flag is to support Safari which has trouble determining function object equality during an animation.
+var isGetComputedStylePatched = false;
 
 var ensureRetickBeforeGetComputedStyle = function() {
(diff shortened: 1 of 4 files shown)


web-animations.js
```diff
@@ -4948,55 +5002,69 @@ Compositor.prototype = {
...
+  if (target.style._webAnimationsStyleInitialised) {
+    return;
+  }
+  try {
     var animatedStyle = new AnimatedCSSStyleDeclaration(target);
     Object.defineProperty(target, 'style', configureDescriptor({
       get: function() { return animatedStyle; }
     }));
   }
+  catch (error) {
```
dstoc: catch should be on the previous line: 
try {
} catch (...

dstoc:
lgtm

web-animations.js

PR #503, opened 2013-12-03
Travis build is now green again

mithro:
- Fixing a Travis problems.
- Fixing tests failing because chrome has fixed things.
- Fixing the error messages so its easier to figure out what is going wrong.


alancutter:
Huzzah, green trees and subpixel font sizing. :)

PR #510, opened 2013-12-11
Update Player to match new spec behaviour and API

alancutter:
This change brings the polyfill up to speed with the spec with regard to Players.
http://dev.w3.org/fxtf/web-animations/#players
http://dev.w3.org/fxtf/web-animations/#the-player-interface

The changes involved are:
- Making the currentTime of a player bounded to the duration of its source content.
- Adding the following to the Player API: playing, cancel(), finish(), play(), pause(), reverse().
- Making the paused attribute on Players readonly.
- Updating tests to the new behaviour.


alancutter:
This change is cancelled in favour of #543.

PR #511, opened 2013-12-11
Create cache buster at test-runner level and pass it down.

mithro:
This means that 100 unique copies of web-animation.js (and similar files) are
not loaded when running inside the test-runner (a big performance problem for
Android).

### DIFF
--- a/test/test-runner.html
+++ b/test/test-runner.html
@@ -164,6 +164,10 @@ function trueOffsetTop(element) {
   }
 }
 
+/**
+ * Get a value for busting the cache.
+ */
+var cacheBuster = '' + window.Date.now();
 
 /**
  * Get the most accurate version of time possible.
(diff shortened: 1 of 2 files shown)


mithro:
@alancutter please take a look as you added the original cache busting code.

test/test-runner.html
```diff
@@ -164,6 +164,14 @@ <h1 id=status>Pending</h1>
   }
 }
 
+/**
+ * Get a value for busting the cache.
+ */
+var cacheBusterValue = '' + window.Date.now();
+function getCacheBuster() {
+  return cacheBusterValue;
+}
```
alancutter: Do we need this function?

mithro: I guess not, originally this function did some other work that I removed.

alancutter:
lgtm

PR #513, opened 2013-12-11
Fix motion path effect when sampling at end of interval

birtles:
I was getting funny effects when hitting the end of the interval and errors in the console.


dstoc:
@alancutter should review this.

It should be pretty simple to write a unit test for this, see eg.

```
test/testcases/unit-test-modify-timing-params.html 
```

shans:
ping @alancutter. @birtles, did you want to write a test case too?

PR #520, opened 2013-12-16
Run tests on Chrome for Android on Linux systems using emulator

mithro:
(Also run on travis as gating browser.)


.travis-run.sh
```diff
@@ -21,5 +21,9 @@ else
     exit 0
   fi
 
-  echo ./run-tests.sh $ARGS | bash || exit 1
+  if [ x$BROWSER == "xAndroid-Chrome" ]; then
```
mikelawther, line 24: I know this isn't part of this review, but I needed to expand the whole file to try and understand the diff. Given that line 17 exits unconditionally, it seems unneeded to have an else on line 18.

.travis-run.sh
```diff
@@ -21,5 +21,9 @@ else
...
+    echo ./run-tests-android.sh $ARGS | bash || exit 1
```
mikelawther, line 25: I haven't seen this echo script.sh $ARGS | bash construct before. Does bash -c "./run-tests-android.sh $ARGS" work here?

.gitmodules
```diff
@@ -12,4 +12,4 @@
 	url = https://github.com/ariya/esprima
 [submodule "test/browserdetect"]
 	path = test/browserdetect
-	url = https://github.com/ded/bowser.git
+	url = https://github.com/web-animations/bowser.git
```
mikelawther, line 15: I don't see any bowser related changes in this CL?

mithro, line 15: They are in the bowser repository and already pushed there. Currently have a pull request upstream for the changes. (https://github.com/ded/bowser/pull/34)

.travis-run.sh
mithro, line 25: bash -c format breaks the quoting :/

test/test-runner.html
```diff
@@ -570,7 +574,7 @@ <h1 id=status>Pending</h1>
   // Deal with tests which are taking too long...
   for (var i in testStates['RUNNING']) {
     var test = testStates['RUNNING'][i];
-    if (now() - test.start > 30e3) {
+    if (now() - test.start > 300e3) {
```
mikelawther, line 577: This seems like a long time. Is it because we're emulating Chrome on Android?

mithro, line 577: Two primary reasons;
- We are using an emulator for Android
- Android Chrome is _much_ slower then desktop chrome on the polyfill (even on real devices).

The longest test goes from less then 4 seconds to taking around 90 seconds...

.travis-run.sh
mikelawther, line 25: Boo :(

mithro, line 25: Yeah, bash quoting/unquoting and environment variables is super hard :/

.gitmodules
mikelawther, line 15: OK.

test/test-runner.html
mikelawther, line 577: Not for this patch, but it'd be nice if we didn't make other faster systems timeout at the speed of the slowest system.

mithro, line 577: Even better if we made the slowest system faster :)

mikelawther:
lgtm.

PR #522, opened 2013-12-17
[DON'T MERGE] Rework property getting code in _assert_style_element.

mithro:
- Makes the code significantly easier to read.
- Reduces duplicate code.
- Runs through get_computed_style better.
# Don't Merge - sending for feedback


mithro:
Change `_set_get_property` to `_set_property` and create `_get_property` (which calls `_set_property` with nothing to set).

Fix the doc strings.

PR #543, opened 2014-02-11
Update Player timing model and API to match spec

alancutter:
http://dev.w3.org/fxtf/web-animations/#players
http://dev.w3.org/fxtf/web-animations/#the-player-interface

Player currentTime is now bounded by the source content unless seeked beyond the boundary.
Players can become held to a certain currentTime value.
The Player API has been updated to reflect the new timing behaviour.

This change depends on #536 landing first.


alancutter:
The tests currently fail due to the dependency on #536, without that change test-player-getcurrent and test-player-cleanup will fail.

shans:
I'll have a more detailed look later, but this basically looks pretty good. Couple of comments:
(1) it'd be easier to review with the unrelated stylistic changes in a separate patch :)
(2) is there any way to kick travis into running again? I've merged #536

alancutter:
1. Which change are you referring to in particular?
2. Done. All is passing now. :)

shans:
ca -> normal, etc. mainly. It took me quite a while to get to the meat of the change.

test/testcases/test-player-cleanup.html
```diff
@@ -70,8 +70,8 @@
 			"Only restarted animation current");
 		assert_equals(_WebAnimationsTestingUtilities._knownPlayers.length, 2,
 			"Restarted and forward filling animations known");
-		assert_styles('#anim', [{"backgroundColor": 'rgb(255, 128, 128)'}], 
-			"takes effect immediately");		
+		assert_styles('#anim', [{"backgroundColor": 'rgb(255, 64, 64)'}],
```
shans, line 73: why did this change?

web-animations.js
```diff
@@ -307,62 +312,108 @@ Player.prototype = {
     try {
       this._currentTime = currentTime;
     } finally {
-      exitModifyCurrentAnimationState(
-          this._hasTicked || this.startTime + this._timeDrift <= lastTickTime);
+      exitModifyCurrentAnimationState(this._hasTicked ||
+          this.startTime + this._storedTimeLag <= lastTickTime);
```
shans: should this test & the one below take holdTimeLag into account too?

test/testcases/test-player-cleanup.html
alancutter, line 73: At 3s player.currentTime is set to 0.
At 5s when the source duration is extended to 4s from 1s the player is only a quarter of the way through (1s/4s) instead of halfway (2s/4s) due to the new bounding behaviour.

web-animations.js
alancutter: Good point, it should probably be startTime + timeLag. I believe what this is checking for is whether the modified state has caused the player's source to become in effect at lastTickTime in which case the expression seems vastly too simplified.
dstockwell can you clarify this?

alancutter: #544 simplifies this logic.

alancutter:
#544 has landed in master, the exitModifyCurrentAnimationState() calls are now unaffected by this change. Please continue review.

shans:
Good to merge. GitHub doesn't want to play nice with unstable Chrome, so you'll need to merge it yourself.

PR #544, opened 2014-02-12
Apply animation effects as soon as an animation is played

alancutter:
This change forces animation effects to apply themselves within the same script context that attached them to the document timeline.

Illustration:
// getComputedStyle(element).color === 'black';
element.animate([{color: 'green'}, {color: 'white'}], 1);
// getComputedStyle(element).color === 'green';

This simplifies some of the logic used when calling exitModifyCurrentAnimationState(). In almost all circumstances the player has undergone a tick before any animation state modification occurs.

### DIFF
--- a/test/testcases/test-update-state.html
+++ b/test/testcases/test-update-state.html
@@ -34,8 +34,20 @@ var expected_failures = {
 var element = document.createElement('span');
 document.body.appendChild(element);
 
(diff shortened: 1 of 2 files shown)


alancutter:
@rjwright

rjwright:
lgtm :)

Issue #502, opened 2013-11-28
Players in reverse don't trigger updates when entering their active interval


alancutter:
http://jsfiddle.net/Nz8S9/2/

PR #545, opened 2014-02-12
Reversed players are not finished when past their source duration

alancutter:
Fix for bug #502.
Reversed players were being identified as finished when their current time was beyond their source content. This resulted in the next tick not being scheduled.
This change fixes this behaviour and takes the sign of the player's playback rate into account when determining whether it will enter its active phase in the future or not.


alancutter:
@shans

PR #546, opened 2014-02-13
Only show passing test results when --verbose is specified

alancutter:
The output of run-tests.sh is quite verbose spending one line for every test case that passes.
This change hides passing output unless --verbose is specified.

See Travis build log for this pull request for an example:
https://travis-ci.org/web-animations/web-animations-js/jobs/18785401


alancutter:
@ericwilligers

ericwilligers:
lgtm

PR #550, opened 2014-02-14
Remove redundant install of the Chrome APK on Android

### LINKED ISSUE #548: Android tests no longer running on Travis
The Android test fails to start in Travis.
https://travis-ci.org/web-animations/web-animations-js

```
$ ./.travis-run.sh
+'[' x == xcheck ']'
+'[' xAndroid-Chrome == xAndroid-Chrome ']'
+echo ./run-tests-android.sh
+bash
Android emulator running at 9140
Waiting for emulator to boot... Booted? 1 Animation? stopped
Redirecting localhost
689 KB/s (27386089 bytes in 38.811s)
    pkg: /data/local/tmp/ChromiumTestShell.apk
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

alancutter:
This fixes #548.

### DIFF
--- a/tools/android/setup.sh
+++ b/tools/android/setup.sh
@@ -101,13 +101,18 @@ else
   CHROME_ACT=.ChromiumTestShellActivity
 fi
 
-$ADB install $CHROME_APK
+function start_chrome () {
+  $ADB shell am start -a android.intent.action.MAIN -n $CHROME_APP/$CHROME_ACT -W
+}
+
+if start_chrome | grep -q "does not exist"; then
+  $ADB install $CHROME_APK
+fi
 
 # Check the chrome binary actually starts without segfaulting
-$ADB shell input keyevent 82   # Send the menu key to unlock the screen
-$ADB shell am start -a android.intent.action.MAIN -n $CHROME_APP/$CHROME_ACT -W  # Start chrome
+start_chrome
 sleep 2
-if $ADB shell am start -a android.intent.action.MAIN -n $CHROME_APP/$CHROME_ACT -W | grep -q "Activity not started, its current task"; then
+if start_chrome | grep -q "its current task has been brought to the front"; then
   echo "Chrome seems to have started okay."
 else
   echo "Chrome seems to have crashed!"


mithro:
Looking at this, I don't think this is the correct solution. We definitely want to run the android setup stuff to run as part of the travis setup step. Doing this will make it happen during the run phase.

I think the correct solution is to check for travis in run-tests-android.sh and not source tools/android/setup.sh in run-tests-android.sh.

Maybe something like

``` bash
if [ x$TRAVIS_JOB_ID == x ]; then
 source tools/android/setup.sh  
fi
```

??

alancutter:
Is the Android setup something we expect to run on dev computers to test Android or only on Travis?

mithro:
You should just be able to ./run-tests-android.sh and have the tests run inside the android simulator (IE it should just download the dependencies). (Currently this only works / tested on a Ubuntu Precise (12.04) machine.)

This is similar how to the ./run-tests.sh for the non-android tests also does all the magical set up.

alancutter:
Okay, a special code path for Travis sounds reasonable in that regard.

mithro:
Yeah. Maybe there should be a generic "SKIP_SETUP" or similar thing that travis sets?

alancutter:
It's not so easy to share environment variables between .travis-setup.sh and .travis-run.sh due to them being sibling execution calls. Instead this change now checks whether it spawned a new emulator or not and only install the APK if it's new.

mithro:
I'm wondering if the APK download/config stuff should be moved to the same section as the rest of the downloading; IE this part;

``` bash
# Download the android emulator and adb
```

Then you can move the installing the Chrome application to the BOOTED/BOOTANIM if?

``` bash
  if [ x$BOOTED == x1 -a x$BOOTANIM == xstopped ]; then

    # Make localhost refer to the host machine, not the emulator.
    # See http://developer.android.com/tools/devices/emulator.html#emulatornetworking
    echo "Redirecting localhost"
    $ADB shell mount -o remount,rw -t yaffs2 /dev/block/mtdblock0 /system
    $ADB shell echo "10.0.2.2 localhost" \> /etc/hosts

    $ADB install $CHROME_APK
    break
```

Actually detecting that the chrome application is installed could possible done with the following;

``` bash
# Check chrome is installed
if ! $ADB shell am start -a android.intent.action.MAIN -n $CHROME_APP/$CHROME_ACT -W | grep XXXPutTheRightThingHereXXX; then
    $ADB install $CHROME_APK
fi
```

alancutter:
Checking whether it's installed or not is probably the least hacky way to resolve this. Patch updated.

mithro:
This patch LGTM. Merging.

Did you figure out what is causing the failure?

PR #551, opened 2014-02-16
Avoid capturing base values for empty compositable value stacks

alancutter:
Empty compositable value stacks do not require the underlying CSS value when they can just clear existing animation effects.
This is a modification of a change by timloh.


dstoc:
lgtm

alancutter:
This patch has changed, might need a re-review.

dstoc:
stilllgtm

alancutter:
Failing Travis build is due to Android tests timing out in Travis.

PR #555, opened 2014-02-19
Avoid useless isDefinedAndNotNull() check in isEffectCallback() check

alancutter:
Removing this check results in an 8% performance improvement.
**Test:** test/perf/balls-replace-compositing.html
**Device:** Nexus 10
**Without patch:**

```
Average: 1.4581406928090335 FPS
Stddev: 0.024775606202678754 FPS
Min: 1.4187670245704818 FPS
Max: 1.5102616996442075 FPS
Median: 1.454523124706587 FPS
```

**With patch:**

```
Average: 1.5757391589373069 FPS
Stddev: 0.013885486397760056 FPS
Min: 1.5620597038094253 FPS
Max: 1.625558592747966 FPS
Median: 1.5697093291425672 FPS
```

### DIFF
--- a/web-animations.js
+++ b/web-animations.js
@@ -1078,9 +1078,7 @@ TimingEvent.prototype = Object.create(window.Event.prototype, {
 });
 
 var isEffectCallback = function(animationEffect) {
-  // TODO: How does WebIDL actually differentiate different callback interfaces?
-  return isDefinedAndNotNull(animationEffect) &&
-      typeof animationEffect === 'function';
+  return typeof animationEffect === 'function';
 };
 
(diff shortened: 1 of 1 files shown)


alancutter:
@mikelawther

mikelawther:
lgtm.

Idle wondering - if there was a global 'debugMode' bool that you could switch on with a url param or something, would V8's JIT be good enough to strip out all the dead code?

alancutter:
Having constants in the language would certainly help that idea.
In the meantime we have an ASSERT_ENABLED flag to short circuit our assertions if it's turned off.

Issue #556, opened 2014-02-20
Polyfill not working in Canary


morethanreal:
The polyfill is not working in today's Chrome Canary (34.0.1847.3 (Official Build 252028) canary) with --enable-experimental-web-platform-features. It works fine without --enable-experimental-web-platform-features and I suspect it has to do with Element.animate landing.

gist: https://gist.github.com/morethanreal/9104830

shans:
document.timeline.play isn't being replaced correctly by the polyfill (we just started providing a native implementation of this). Fix pending.

alancutter:
We really ought to have an experimental Chrome target for Travis to catch these in future.

PR #558, opened 2014-02-21
allow web animations polyfill to work in presence of experimental-web-platform-features flag on canary

shans:
This addresses #556


alancutter:
lgtm

PR #559, opened 2014-02-21
Add unstable experimental Chrome Travis target

alancutter:
We failed to detect a fault in the polyfill when document.timeline gained a native implementation in experimental Chrome, see issue #556.
This patch adds an experimental Chrome target to Travis so we can catch these types of failures in future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
     - BROWSER=Chrome-stable ARGS='-x -b Chrome -u'
     - BROWSER=Chrome-beta ARGS='-x -b Chrome -u'
     - BROWSER=Chrome-unstable ARGS='-x -b Chrome -u'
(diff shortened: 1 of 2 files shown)


mithro:
The --flags option will break if you try and give it arguments which have spaces in it. Can you change the --flags argument to be an append type rather then using split.

```
'append' - This stores a list, and appends each argument value to the list. This is useful to allow an option to be specified multiple times. Example usage:
>>>
>>> parser = argparse.ArgumentParser()
>>> parser.add_argument('--foo', action='append')
>>> parser.parse_args('--foo 1 --foo 2'.split())
Namespace(foo=['1', '2']) 
```

alancutter:
Done.

mithro:
LGTM. Merge if travis goes green.

Issue #557, opened 2014-02-21
chrome_debug.log file no longer exists in Chrome unstable


alancutter:
https://travis-ci.org/web-animations/web-animations-js/jobs/19297487

```
Ran 1481 tests in 41.100s
OK
Traceback (most recent call last):
  File "tools/python/run-tests.py", line 752, in <module>
    shutil.copy(os.path.join(user_data_dir, "chrome_debug.log"), ".")
  File "/usr/lib/python2.7/shutil.py", line 117, in copy
    copyfile(src, dst)
  File "/usr/lib/python2.7/shutil.py", line 82, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 2] No such file or directory: '/tmp/tmpPoZP1H/chrome_debug.log'
+exit 1
The command "./.travis-run.sh" exited with 1.
after_failure
$ for file in *.log; do echo $file; echo "======================"; cat $file; done || true
*.log
======================
cat: *.log: No such file or directory
Done. Your build exited with 1.
```

alancutter:
This comment is a bookmark for this page: http://www.chromium.org/for-testers/enable-logging

alancutter:
--enable-logging broken in Chrome unstable: https://code.google.com/p/chromium/issues/detail?id=345582

PR #560, opened 2014-02-21
Check if Chrome debug log exists before copying

alancutter:
#557

Chrome unstable channel debug logging is broken.
This patch makes run-tests.py more resilient to missing Chrome debug log files.


mithro:
LGTM. Merge if travis goes green.

PR #562, opened 2014-02-25
Support fill mode auto

alancutter:
This change adds support for fill mode auto to the API and makes it the default for all timing inputs.
http://dev.w3.org/fxtf/web-animations/#idl-def-FillMode


alancutter:
@ericwilligers

ericwilligers:
lgtm

PR #565, opened 2014-03-03
I hold this patch to be self-evident

shans:


shans:
@dstockwell want to take a gander?

dstoc:
lgtm, but seems to indicate that we're missing a test

shans:
ptal

PR #569, opened 2014-03-24
element.animate now returns player

ericwilligers:
Calling animate on an element now returns a player.

http://dev.w3.org/fxtf/web-animations/
Section 5.10 The Animatable interface


alancutter:
lgtm, please squash these commits and make sure none of the Travis failures are related to your change before merging.

PR #570, opened 2014-03-24
animate returns player

ewilligers:
Calling animate on an element now returns a player

http://dev.w3.org/fxtf/web-animations/
Section 5.10 The Animatable interface


alancutter:
lgtm.

Issue #566, opened 2014-03-05
Support for transformOrigin and matrix3d


heatherdouglass:
Hello, I have been using the polyfill and it's been working great,  thanks!  But I found support for animating matrix3d transforms and transformOrigin to be missing.  Can support for these be added?  Thanks!

alancutter:
Adding support for matrix3d now. :)

alancutter:
Support for these two added.

PR #573, opened 2014-04-03
Add support for animating 3D transforms

alancutter:
Support animating 3D transform functions.
http://dev.w3.org/csswg/css-transforms/#transform-functions

Bug: #566


test/testcases/auto-test-transform-primitives-checks.js
```diff
@@ -273,75 +273,17 @@ timing_test(function() {
...
-      {'transform':'matrix(1, 0.36397023426620234, 0.8390996311772799, 1, 0, 0)'},
-      {'transform':'matrix(1, 0, 0, 1, 0, 0)'},
-      {'transform':'matrix(1, 0, 0.36397023426620234, 1, 0, 0)'},
-      {'transform':'matrix(1, 0, 0.8390996311772799, 1, 0, 0)'},
-      {'transform':'matrix(1, 0, 0, 1, 0, 0)'},
-      {'transform':'matrix(1, 0.8390996311772799, 0, 1, 0, 0)'},
+      {'transform':'matrix(1, 0.8391, 0, 1, 0, 0)'},
       {'transform':'matrix(1, 0, 0, 1, 0, 0)'},
       {'transform':'matrix(-1, 1, -1, -1, 10, 10)'},
       {'transform':'matrix(1, 0, 0, 1, 0, 0)'},
```
shans, line 289: Copy this test to something-something-something-3d, reinstate the old one. The old test is a guarantee that 2d matrix behaviour doesn't change with the introduction of this patch, and shouldn't be refactored in the same CL as this.

web-animations.js
```diff
@@ -3716,7 +3716,11 @@ function build3DRotationMatcher() {
     for (var i = 0; i < 3; i++) {
       out.push(r[i].px);
     }
-    out.push(r[3]);
+    var angle = 0;
+    for (var unit in r[3]) {
+      angle += convertToDeg(r[3][unit], unit);
```
shans, line 3821: _sigh_ units in matrices. It's a travesty.

web-animations.js
```diff
@@ -3964,28 +3968,149 @@ function dot(v1, v2) {
...
+    a[0] * b[8] + a[4] * b[9] + a[8] * b[10] + a[12] * b[11],
+    a[1] * b[8] + a[5] * b[9] + a[9] * b[10] + a[13] * b[11],
+    a[2] * b[8] + a[6] * b[9] + a[10] * b[10] + a[14] * b[11],
+    a[3] * b[8] + a[7] * b[9] + a[11] * b[10] + a[15] * b[11],
+
+    a[0] * b[12] + a[4] * b[13] + a[8] * b[14] + a[12] * b[15],
+    a[1] * b[12] + a[5] * b[13] + a[9] * b[14] + a[13] * b[15],
+    a[2] * b[12] + a[6] * b[13] + a[10] * b[14] + a[14] * b[15],
+    a[3] * b[12] + a[7] * b[13] + a[11] * b[14] + a[15] * b[15]
+  ];
```
shans, line 4093: I wonder if this is much slower in the 2d case? We should measure some time.

web-animations.js
```diff
@@ -4230,10 +4381,8 @@ var transformType = {
               value[i].d[1] + ', ' + value[i].d[2] + ') ';
           break;
         case 'matrix':
-          out += value[i].t + '(' +
-              n(value[i].d[0]) + ', ' + n(value[i].d[1]) + ', ' +
-              n(value[i].d[2]) + ', ' + n(value[i].d[3]) + ', ' +
-              n(value[i].d[4]) + ', ' + n(value[i].d[5]) + ') ';
+        case 'matrix3d':
+          out += value[i].t + '(' + value[i].d.map(n).join(', ') + ') ';
```
shans, line 4507: It's not 100% clear from the specification, but I _think_ that interpolating two 2d matrices should produce a 2d matrix as CSS output. This may also be the case for 3d matrices that are zero in all the right places.

You should experiment with what Chrome does right now and reproduce that.

web-animations.js
```diff
@@ -3835,10 +3837,12 @@ var decomposeMatrix = (function() {
   }
 
   function decomposeMatrix(matrix) {
```
shans, line 3941: The spec now has 2D and 3D versions of this. We should verify they produce the same result for 2D matrices now; but I think that can happen later.

web-animations.js
alancutter, line 3821: This is the rotate3d function, not a matrix yet. :)

web-animations.js
alancutter, line 4507: I'm not sure what you mean by produce. getComputedStyle() will return a 2d matrix even if a 3d equivalent of a 2d matrix is set.

test/testcases/auto-test-transform-primitives-checks.js
alancutter, line 289: This change adds a fix to skew handling (line 4171) in the matrix decomposition interpolation case which produces slightly different intermediate values for this test's matrix to matrix interpolation. There's no obvious visual difference between them for this test. The other lines affected are due to changes in the test generator since these expectations were captured, now values are rounded to 4 decimal places. 

Ideally this file wouldn't change given that it's only testing 2d transform functions but we use the same matrix decomposition path for 2d and 3d.

shans, line 289: I commented on the wrong test...

test/testcases/auto-test-matrix-transforms-checks.js
```diff
@@ -1,56 +1,160 @@
...
+      {'transform':'matrix3d(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 30, 1)'},
+      {'transform':'matrix3d(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 10, 20, 40, 1)'},
+      {'transform':'none'},
+      {'transform':'none'},
+      {'transform':'none'},
+      {'transform':'none'},
+      {'transform':'matrix(1, 0, 0, 1, 100, 0)'},
+      {'transform':'matrix(0.9397, 0.342, -0.342, 0.9397, 100, 0)'},
+      {'transform':'matrix(0.4184, 0.8972, -0.8972, 0.4184, 100, 0)'},
+      {'transform':'matrix(0.4698, 0.171, -0.171, 0.4698, 100, 0)'},
```
shans: There's no correspondence between the new results and the old ones. You need to fix this.

Please copy your modified test into a new file and give it a new name, then reinstate the old test so we can eyeball that the results haven't changed (or at least haven't changed significantly).

web-animations.js
dstoc, line 4507: Shane says: we should match the result of getComputedStyle in the style we set. Otherwise, when we stack animations, the results may be incorrect (I think).

test/testcases/auto-test-matrix-transforms-checks.js
alancutter: Done.

web-animations.js
alancutter, line 3941: Added TODO.

web-animations.js
alancutter, line 4507: Done, see line 4199.

shans:
lgtm

PR #574, opened 2014-04-03
Add new breaking changes policy

maxh:
Update readme to explain new breaking changes policy.  Not sure about the specifics -- Shane should review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,4 +232,4 @@ For running tests or building minified files, consult the
 
 ## Breaking changes
 
-When we make a potentially breaking change to polyfill's API surface (like a rename) we keep the old version around on another branch, deprecated, for 3 months, and ensure that there are console warnings that a change is pending. We also announce anything that isn't a bug fix on web-animations-changes@googlegroups.com (https://groups.google.com/forum/#!forum/web-animations-changes).
+When we make a potentially breaking change to the polyfill's API surface (like a rename) we keep the old version around, deprecated, for 3 months, and ensure that there are console warnings that a change is pending. We also announce anything that isn't a bug fix on web-animations-changes@googlegroups.com (https://groups.google.com/forum/#!forum/web-animations-changes).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -229,3 +229,7 @@ then the polyfill will provide a console warning in browsers where these feature
 
 For running tests or building minified files, consult the
 [tooling information](http://www.polymer-project.org/resources/tooling-strategy.html).
+
+## Breaking changes
+
+When we make a potentially breaking change to polyfill's API surface (like a rename) we keep the old version around on another branch, deprecated, for 3 months, and ensure that there are console warnings that a change is pending. We also announce anything that isn't a bug fix on web-animations-changes@googlegroups.com (https://groups.google.com/forum/#!forum/web-animations-changes).
```
shans: lose 'on another branch'. We'll keep both versions in head. Otherwise this is good.

maxh:
Done.  I tweaked the wording to make this more clear too.

README.md
```diff
@@ -229,3 +229,7 @@ then the polyfill will provide a console warning in browsers where these feature
 
 For running tests or building minified files, consult the
 [tooling information](http://www.polymer-project.org/resources/tooling-strategy.html).
+
+## Breaking changes
+
+When we make a potentially breaking change to the polyfill's API surface (like a rename) we'll continue supporting the old version, deprecated, for 3 months, and ensure that there are console warnings that a change is pending. We also announce anything that isn't a bug fix on web-animations-changes@googlegroups.com (https://groups.google.com/forum/#!forum/web-animations-changes).
```
alancutter: I wonder if this should be explicit that the 3 month deprecation is automatically handled by the polyfill and the old function will be removed even if you don't update the version of the polyfill. If you see deprecation warnings you can't avoid it by not updating.

maxh: I took a stab at making this more clear.

PR #575, opened 2014-04-08
Rename Timeline to AnimationTimeline

ewilligers:
Rename Timeline to AnimationTimeline
(From Web Animations minutes 27 Feb 2014
https://etherpad.mozilla.org/ep/pad/view/ro.WCBLOuse6D5/latest )

Introduce deprecation reporting, where a deprecated feature continues to
work for three months after deprecation, but a warning is logged to the
console when the feature is first used.

Illustrate how the deprecation reporting system is used, using the
obsolete Timeline constructor.

### DIFF
--- a/web-animations.js
+++ b/web-animations.js
@@ -55,6 +55,7 @@ function createDummyElement() {
 
 var features = detectFeatures();
 var constructorToken = {};
+var deprecationsSilenced = {};
 
 var createObject = function(proto, obj) {
   var newObject = Object.create(proto);
@@ -67,6 +68,24 @@ var createObject = function(proto, obj) {
 
 var abstractMethod = function() {
   throw 'Abstract method not implemented.';
+};
+
+var deprecated = function(name, deprecationDate, advice) {
+  if (deprecationsSilenced[name]) {
+    return;
+  }
+  var today = new Date();
+  var cutoffDate = new Date(deprecationDate);
+  cutoffDate.setMonth(cutoffDate.getMonth() + 3); // 3 months grace period
+
+  if (today < cutoffDate) {
+    console.warn('Web Animations: ' + name +
+        ' is deprecated and will stop working on ' +
+        cutoffDate.toDateString() + '. ' + advice);
+    deprecationsSilenced[name] = true;
+  } else {
+    throw new Error(name + ' is no longer supported. ' + advice);
+  }
 };
 
 var IndexSizeError = function(message) {
@@ -193,7 +212,7 @@ var isDefinedAndNotNull = function(val) {
 
 
 /** @constructor */
-var Timeline = function(token) {
+var AnimationTimeline = function(token) {
   if (token !== constructorToken) {
     throw new TypeError('Illegal constructor');
   }
@@ -204,7 +223,7 @@ var Timeline = function(token) {
   }
 };
 
-Timeline.prototype = {
+AnimationTimeline.prototype = {
   get currentTime() {
     if (this._startTime === undefined) {
       this._startTime = documentTimeZeroAsClockTime;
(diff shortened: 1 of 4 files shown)


web-animations.js
```diff
@@ -242,6 +258,11 @@ Timeline.prototype = {
   }
 };
 
+var Timeline = function(token) {
+  reportDeprecated('Timeline', '2014-04-08', 'Use AnimationTimeline.');
+  return new AnimationTimeline(token);
+};
+
```
alancutter: We never actually exposed constructing Timelines in the API however we did expose the classname.
You'll probably want to show the deprecation message in the getter for 'Timeline' on the window object example:
Object.defineProperty(window, 'Timeline', {get: function() {reportDeprecated(...);}, ...});
Don't forget to check on all browsers that this type of modification of the root object doesn't crash the polyfill.

web-animations.js
```diff
@@ -69,6 +70,21 @@ var abstractMethod = function() {
...
 };
 
+var reportDeprecated = function(name, deprecationDate, advice) {
+  var today = new Date();
+  var cutoffDate = new Date(deprecationDate);
+  cutoffDate.setMonth(cutoffDate.getMonth() + 3); // 3 months grace period
+
+  if (today < cutoffDate) {
+    if (!deprecationsSilenced[name]) {
+      console.log('Warning: ' + name + ' has been deprecated and will stop working on ' + cutoffDate.toDateString() + '. ' + advice);
```
alancutter: Use console.warn().
I wonder if we should be explicit in the message that it's coming from the Web Animations polyfill.
In the case of a rename we should direct the user on what the right thing to use is. Consider using this as a template (based on the Chrome deprecation messages):
console.warn('Web Animations: ' + name + ' is deprecated and will stop working on ' + date + '. Please use ' + alternative + ' instead.');

web-animations.js
```diff
@@ -5340,4 +5360,17 @@ window._WebAnimationsTestingUtilities = {
   _pacedTimingFunction: PacedTimingFunction
 };
 
+// Timeline is deprecated
+Object.defineProperty(window, 'Timeline', {
+  get : function() {
+    reportDeprecated('Timeline', '2014-04-08', 'AnimationTimeline');
+    return AnimationTimeline;
+  },
+  set : function(newValue) {
```
alancutter: Unnecessary space after get and set.

alancutter:
lgtm with nit.

web-animations.js
```diff
@@ -69,6 +70,27 @@ var abstractMethod = function() {
   throw 'Abstract method not implemented.';
 };
 
+var reportDeprecated = function(name, deprecationDate, alternative) {
+  var today = new Date();
```
shans, line 77: guard against deprecationsSilenced first. This is the most likely condition.

web-animations.js
```diff
@@ -69,6 +70,27 @@ var abstractMethod = function() {
...
+  var cutoffDate = new Date(deprecationDate);
+  cutoffDate.setMonth(cutoffDate.getMonth() + 3); // 3 months grace period
+  var advice = '';
+  if (alternative) {
+    advice = 'Please use ' + alternative + ' instead.';
+  }
```
shans: Although it's slightly more verbose, I'd rather pass advice in directly, so we can do things like:

'Web Animations: Timeline.onstart is deprecated and will stop working on &lt;whatever&gt;. See &lt;explainer doc&gt; about how to recast Timeline events as custom effect callbacks'

web-animations.js
```diff
@@ -69,6 +70,27 @@ var abstractMethod = function() {
   throw 'Abstract method not implemented.';
 };
 
+var reportDeprecated = function(name, deprecationDate, alternative) {
```
shans: 'deprecated' instead of 'reportDeprecated'?

web-animations.js
```diff
@@ -5340,4 +5362,17 @@ window._WebAnimationsTestingUtilities = {
   _pacedTimingFunction: PacedTimingFunction
 };
 
+// Timeline is deprecated
+Object.defineProperty(window, 'Timeline', {
+  get: function() {
+    reportDeprecated('Timeline', '2014-04-08', 'AnimationTimeline');
+    return AnimationTimeline;
+  },
+  set: function(newValue) {
```
shans: readonly Timeline is fine.

PR #576, opened 2014-04-09
Support animating transform-origin

alancutter:
Add support for animating the transform-origin property.

### DIFF
--- a/web-animations.js
+++ b/web-animations.js
@@ -5413,7 +5413,8 @@ window._WebAnimationsTestingUtilities = {
   _hsl2rgb: hsl2rgb,
   _types: propertyTypes,
   _knownPlayers: PLAYERS,
-  _pacedTimingFunction: PacedTimingFunction
+  _pacedTimingFunction: PacedTimingFunction,
+  _prefixProperty: prefixProperty
 };
 
 })();
--- a/web-animations.js
+++ b/web-animations.js
@@ -2923,7 +2923,7 @@ var positionType = {
     return value.map(percentLengthType.toCssValue).join(' ');
   },
   fromCssValue: function(value) {
-    var tokens = positionType.consumeTokensFromString(value);
+    var tokens = positionType.consumeAllTokensFromString(value);
     if (!tokens || tokens.length > 4) {
       return undefined;
     }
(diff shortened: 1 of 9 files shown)


web-animations.js
```diff
@@ -2984,6 +2984,18 @@ var positionType = {
     }
     return out.every(isDefinedAndNotNull) ? out : undefined;
   },
+  consumeTokensFromString: function(remaining) {
```
shans: can you call this consumeAllTokensFromString? I think the name is too similar to consumeTokenFromString atm.

web-animations.js
```diff
@@ -2911,21 +2923,9 @@ var positionType = {
...
-      tokens.push(result.value);
-      remaining = result.remaining;
-      if (!result.remaining.trim()) {
-        break;
-      }
-      if (tokens.length >= 4) {
-        return undefined;
-      }
+    var tokens = positionType.consumeTokensFromString(value);
+    if (!tokens || tokens.length > 4) {
```
shans, line 2927: &gt;= 4?

web-animations.js
```diff
@@ -3105,6 +3117,77 @@ var rectangleType = {
...
+      percentLengthType.interpolate(from[2], to[2], f)
+    ];
+  },
+  toCssValue: function(value) {
+    var result = percentLengthType.toCssValue(value[0]) + ' ' +
+        percentLengthType.toCssValue(value[1]);
+    for (var unit in value[2]) {
+      if (value[2][unit] !== 0) {
+        return result + ' ' + percentLengthType.toCssValue(value[2]);
+      }
```
shans, line 3143: does this handle [{'%': 0}, {'%': 0}, {px: 10, '%': 10}]?

web-animations.js
```diff
@@ -3105,6 +3117,77 @@ var rectangleType = {
...
+          out[0] = tokens[0];
+          out[1] = tokens[1];
+        } else if (positionType.isVerticalToken(tokens[0]) &&
+            positionType.isHorizontalToken(tokens[1])) {
+          out[0] = tokens[1];
+          out[1] = tokens[0];
+        } else {
+          return undefined;
+        }
+        if (tokens.length === 3) {
```
shans: You could avoid this test by doing

case 3:
  var out = ...
  // this stuff
case 2:
  // the rest

web-animations.js
alancutter, line 2927: The check is different because it's not in the same place in relation to the while (true) and the break. Using >= 4 breaks existing unit tests.

web-animations.js
alancutter, line 3143: Added comment for clarity.

web-animations.js

web-animations.js

shans:
lgtm

alancutter:
This change is for issue #566.

PR #581, opened 2014-04-11
Change expected test failures formatting

alancutter:
This patch changes the way we specify expected test failures for particular browsers.
This change is necessary to support targeting Chrome but not Chrome on Android for expected test failures.


ewilligers:
lgtm

PR #582, opened 2014-04-16
Rename TimedItem.specified to TimedItem.timing

ewilligers:
TimedItem.specified is now deprecated. It has been replaced by
TimedItem.timing.

Meeting minutes describing this change:
http://lists.w3.org/Archives/Public/public-fx/2014AprJun/0005.html


alancutter:
lgtm, don't forget to send out the email and create a removal calendar entry.

PR #584, opened 2014-04-22
Change from seconds to milliseconds as the unit of time

dstoc:


dstoc:
This should not be merged until messages have been sent about this breaking change.

test/testcases/auto-test-transform-functions-checks.js
```diff
@@ -103,7 +103,7 @@ timing_test(function() {
       {'transform':'matrix(0.7386, 0.1302, -0.1302, 0.7386, 100, 0)'},
     ]);
   });
-  // This check is equivalent to checking at t=1s, it is placed all the way out here to allow manual viewing
+  // This check is equivalent to checking at t=1s, it ims placed all the way out here to allow manual viewing
```
alancutter: Prefixed the wrong s!

alancutter:
I think the README needs updating too.

alancutter:
I'm not liking how all the tests operate in the order of milliseconds now, though I can understand it's probably more or less okay. Is it particularly challenging to convert them all over to their equivalent millisecond values?

dstoc:
> I'm not liking how all the tests operate in the order of milliseconds now, though I can understand it's probably more or less okay. Is it particularly challenging to convert them all over to their equivalent millisecond values?

I'll have a go.

dstoc:
> I'm not liking how all the tests operate in the order of milliseconds now, though I can understand it's probably more or less okay. Is it particularly challenging to convert them all over to their equivalent millisecond values?
> I'll have a go.

Now using original timings for most auto tests, not as important for most unit- and test- tests which are harder to convert.

alancutter:
Thanks!
lgtm

PR #589, opened 2014-04-30
fix: undefined onfinish event target

douglasduteil:
Hi it's me again

When listening to the onfinish event, the target of the created event is _undefined_.
Is it supposed to be the AnimationPlayer source there ?

### DIFF
--- a/test/testcases/test-player-finish-event.html
+++ b/test/testcases/test-player-finish-event.html
@@ -35,6 +35,7 @@ function assertFinishEvents(description, player, expectedEvents) {
         asyncTestHandle.step(function() {
           assert_equals(event.currentTime, expectedEvent.currentTime, 'event.currentTime');
           assert_equals(event.timelineTime, expectedEvent.timelineTime, 'event.timelineTime');
+          assert_equals(event.target, player, 'event.target');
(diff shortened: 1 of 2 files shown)


dstoc:
> Is it supposed to be the AnimationPlayer source there ?

It should be the AnimationPlayer itself. Could you also update test/testcases/test-player-finish-event.html to assert this?

douglasduteil:
@dstockwell Does it suit you ?

test/testcases/test-player-finish-event.html
```diff
@@ -35,6 +35,7 @@
         asyncTestHandle.step(function() {
           assert_equals(event.currentTime, expectedEvent.currentTime, 'event.currentTime');
           assert_equals(event.timelineTime, expectedEvent.timelineTime, 'event.timelineTime');
+          assert_equals(event.target, expectedEvent.target || player, 'event.target');
```
alancutter: It's probably fine to just use player here instead of expectedEvent.target which is never used.

alancutter:
lgtm, looking into the failing Travis build.

test/testcases/test-player-finish-event.html
douglasduteil: Alright

alancutter:
Thanks for submitting the fix!

Issue #590, opened 2014-05-20
experimental branch errors out on backgroundImage


heatherdouglass:
Hello,
If I create an animation that animates backgroundImage and then try to use the experimental `effectValueAtLocalTime` API I get an error, `TypeError: Cannot read property 'toCssValue' of undefined` at line 1821.  If I simply play the animation though then it does work.  Can the API be fixed for backgroundImage?  Thanks.
~Heather~

shans:
Fix coming in #591

shans:
Fixed in experimental and experimental-rebase

heatherdouglass:
Hello,
Thanks for the quick turn around, but with the new code I am still seeing the same error, 'Uncaught TypeError: Cannot call method 'toCssValue' of undefined' at line 1821.
Thanks,
~Heather~

shans:
Hi Heather,

Can you provide a test case?

Thanks,
    -Shane

shans:
Sounds like this is fixed now.

PR #591, opened 2014-05-21
Ensure that non-numeric property types work with effectValueAtLocalTime.

shans:
Non-numeric properties don't have specific entries in the propertyTypes
dictionary. Rather than looking properties up directly, using getType
ensures the generic type entry is used when none other is available.

### DIFF
--- a/web-animations.js
+++ b/web-animations.js
@@ -1643,7 +1643,8 @@ function toUsableValue(property, value) {
       }
       var propertyType = getType(property);
       var internalUnderlying = propertyType.fromCssValue(underlying);
-      return propertyType.toCssValue(value.compositeOnto(property, internalUnderlying));
+      return propertyType.toCssValue(value.compositeOnto(property,
+	  internalUnderlying));
     }
   }   
 }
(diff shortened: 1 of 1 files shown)


shans:
This is a fix for #590

web-animations.js
```diff
@@ -1641,8 +1641,9 @@ function toUsableValue(property, value) {
       if (underlying == undefined) {
         throw new TypeError('Underlying value undefined');
       }
-      var internalUnderlying = propertyTypes[property].fromCssValue(underlying);
-      return propertyTypes[property].toCssValue(value.compositeOnto(property, internalUnderlying));
+      var propertyType = getType(property);
+      var internalUnderlying = propertyType.fromCssValue(underlying);
+      return propertyType.toCssValue(value.compositeOnto(property, internalUnderlying));
```
dstoc: This line is still too long :P

test/browserdetect
```diff
@@ -1 +1 @@
-Subproject commit b6a2f65cc857c968067440d8bfb8526291286b15
+Subproject commit 23545d440337aed62652c1c2c56cce99d40d21bf
```
dstoc, line 1: wat?

dstoc:
lgtm

PR #593, opened 2014-05-26
Update test failure expectations for Firefox 31

alancutter:


test/testcases/auto-test-path.html
```diff
@@ -106,7 +106,7 @@
     ],
     message: 'Doesn\'t quite follow path correctly.',
   }, {
-    browser_configurations: [{ firefox: true, version: '28|29|30' }],
+    browser_configurations: [{ firefox: true, version: '28|29|30|31' }],
```
dstoc: Are there any versions of ff where these expectations aren't used? Do we still test < 28?

alancutter: Removed untested Firefox versions.

alancutter:
dstockwell: Poke.

dstoc:
lgtm

PR #595, opened 2014-05-27
More fixes to use getType rather than directly look up the propertyTypes dictionary.

shans:


shans:
I'm not 100% sure whether this fixes the remaining issues in #590.

dstoc:
lgtm

PR #599, opened 2014-05-28
Remove accumulate as it is no longer part of the specification.

shans:


dstoc:
lgtm

dstoc:
Travis says:

> Line 4886, E:0007: Should have 3 blank lines before a constructor/interface.

shans:
Buh? That is _not_ part of the Google Style guide, as far as I can tell. Where on earth did it come from?

It's also a stupid rule :P

PR #601, opened 2014-06-03
Removing --strict so lint only enforces "Google Javascript Style" rather then "Closure Javascript Style".

mithro:
@shans


dstoc:
lgtm

PR #602, opened 2014-06-12
Animate SVG height and circle coordinates

ewilligers:
Height was already being animated in CSS animations but not in SVG
animations.


alancutter:
lgtm, need to fix the check failure.

PR #603, opened 2014-06-12
Make propertyIsSVGAttrib available to test harness.

ericwilligers:
We were previously specifying the list of SVG attributes in 3 files. Now we only need to specify the list once.


alancutter:
lgtm

PR #604, opened 2014-06-12
Interpolate SVG colors

ewilligers:
We were previously discretely stepping between SVG colors instead of
interpolating.

### DIFF
--- a/web-animations.js
+++ b/web-animations.js
@@ -4570,6 +4570,8 @@ var propertyTypes = {
   color: colorType,
   cx: lengthType,
   cy: lengthType,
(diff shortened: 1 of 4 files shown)


web-animations.js
```diff
@@ -4607,6 +4610,8 @@ var propertyTypes = {
   perspectiveOrigin: originType,
   r: lengthType,
   right: percentLengthAutoType,
+  stopColor: colorType,
```
alancutter, line 4613: It shouldn't be too hard to add these other properties to the same test.

alancutter:
lgtm

Issue #605, opened 2014-06-12
SVG transform animations fills forwards


ewilligers:
In the following example, the scale remains in effect after the 2 seconds have passed.

```
  // We have a polyfill bug here: fill 'none' is not being respected.
  element.animate([{transform: 'scale(1)'}, {transform: 'scale(3)'}], {duration: 2000, fill: 'none'});
```

This observed faulty behavior can be avoided by noting that transform is an SVG attribute; it must be added to svgProperties.

ewilligers:
My commit was [reverted](https://github.com/web-animations/web-animations-js/pull/609) as it caused a Travis failure in Firefox.

PR #606, opened 2014-06-12
SVG transforms now complete when fill=none

ewilligers:
The transforms previously remained in effect, as if the page author had
requested fill=forwards

https://github.com/web-animations/web-animations-js/issues/605

### DIFF
--- a/test/testcases/auto-test-svg-transform.html
+++ b/test/testcases/auto-test-svg-transform.html
@@ -1,0 +1,39 @@
+<!--
+Copyright 2012 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


test/testcases/auto-test-svg-transform.html
```diff
@@ -0,0 +1,39 @@
...
+}
+
+.expectation {
+  fill: red;
+}
+</style>
+<svg width="600px" height="200px">
+  <rect class="expectation" x="190px" y="60px" width="120px" height="120px">
+  </rect>
+  <rect class="anim" x="10px" y="10px" width="20px" height="20px" transform="none"></rect>
```
ewilligers: Without transform=none, the animation is still observed to not fill forwards, but the tests fail with transform='null' at times 1_1000 and 1.2_1000

alancutter:
lgtm

PR #611, opened 2014-06-23
Typos in README; fill is forwards and backwards

rictic:
... not 'forward' and 'backward'. This had me puzzled for like an hour.

cite from the spec [section 3.7.1 Fill Modes](http://dev.w3.org/fxtf/web-animations/#fill-modes)


alancutter:
Thanks for submitting the fix!

Issue #614, opened 2014-06-25
EffectCallback signature doesn't match spec


morethanreal:
According to spec the params are timeFraction, currentTarget, animation but the polyfill calls it with timeFraction, animation, previousTimeFraction.

alancutter:
Thanks for pointing that out, should be fixed now.

PR #615, opened 2014-06-26
Update EffectCallback parameters to match spec

alancutter:
#614

The parameters passed to custom effect callbacks should be:
`callback EffectCallback = void (double? timeFraction, Animatable currentTarget, Animation animation);`
Spec: http://dev.w3.org/fxtf/web-animations/#idl-def-EffectCallback


dstoc:
lgtm

PR #618, opened 2014-06-27
[RFC] Decompose matrices individually to allow rotate(0deg) -> rotate(360deg) after a matrix

joeytwiddle:
Currently in `transformType.interpolate()` the presence of a matrix forces itself and _all later transforms_ to be decomposed into their components for interpolation.  But decomposing `rotate(0deg)` and `rotate(360deg)` into matrices breaks the interpolation of the angle value.  (Because both of those rotations decompose to a rotation of 0!)

This patch attempts to fix this: when a matrix (or incompatible types) are encountered, we only perform decomposition at that stage of the composition.  Later interpolations can be performed without decomposition into matrices.

I added a test case `auto-test-composite-transforms-matrix-rotate.html` (a slight modification of `
auto-test-composite-transforms.html`).

```
dt.play(new Animation(divs[0], [
  {transform: "matrix(1,0,0,1,0,0) rotate(0deg)"},
  {transform: "matrix(1,0,0,1,200,0) rotate(360deg)"},
], timing));
```

With the old behaviour the first two divs do not rotate at all.  With the new behaviour you will see them perform a full rotation.

Concerns:
- It looks like `from.decompositionPair = to;` and `from.decomposition` were assigned as a way of caching the decomposed matrix, but I did not reproduce caching in this patch.

Alternatives:
- I thought a workaround might be to specify additional keyframes with intermediate angles, e.g. `0deg`, `120deg`, `240deg` and `360deg` but the decomposition still causes 240 to rotate anti-clockwise to "360" (because it was decomposed to 0).

### DIFF
--- a/test/testcases/auto-test-composite-transforms-matrix-rotate.html
+++ b/test/testcases/auto-test-composite-transforms-matrix-rotate.html
@@ -49,11 +49,11 @@ limitations under the License.
 
 </style>
 
-<div>Right edge of each box should align with black line at end of test.</div>
+<div>The first two divs should rotate 360 degrees.</div>
 <div id="a" class="anim"></div>
 <div id="a" style="top: 50px; left: 200px;" class="expected"></div>
 <div id="b" class="anim"></div>
-<div id="b" style="top: 150px; left: 0px;" class="expected"></div>
+<div id="b" style="top: 150px; left: 100px;" class="expected"></div>
 <div id="c" class="anim"></div>
 <div id="c" style="top: 250px; left: 200px;" class="expected"></div>
 <div id="d" class="anim"></div>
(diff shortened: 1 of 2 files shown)


shans:
I think that this patch has the _correct_ behavior for the example you've provided:

```
dt.play(new Animation(divs[0], [
  {transform: 'matrix(1,0,0,1,0,0) rotate(0deg)'},
  {transform: 'matrix(1,0,0,1,200,0) rotate(360deg)'},
], timing));
```

but _incorrect_ behavior for cases such as:

```
dt.play(new Animation(divs[0], [
  {transform: 'rotate(45deg) rotate(0deg)'},
  {transform: 'translate(100px) rotate(360deg)'},
], timing));
```

This is because the former case is an example of matched transform components, which should interpolate component-wise, but the latter is an example of mismatched transform components, which should fall back on matrix interpolation from the first mismatch.

joeytwiddle:
Edit: _"which should fall back on matrix interpolation from the first mismatch"_  So asymmetry **_is**_ part of the spec.  I think this answers my striked question below.

&lt;verbose overly=true&gt;<strike>My patch would interpolate the `rotate(45deg)` with the `translate(100px)` and then move on to interpolate the next pair.</strike>

<strike>The original behaviour would (due to the mismatch) collapse the `rotate(45deg) rotate(0deg)` into a matrix, and likewise for the other keyframe, then interpolate the decomposed matrices.</strike>

<strike>Which do you think is the correct behaviour?</strike>

<strike>My concern is that with the original behaviour, this would act "asymmetrically":</strike>

```
dt.play(new Animation(divs[0], [
  {transform: 'rotate(90deg)  rotate(45deg)    rotate(0deg)'},
  {transform: 'rotate(180deg) translate(100px) rotate(360deg)'},
], timing));
```

<strike>The first pair `rotate(90deg)` to `rotate(180deg)` would be interpolated, but the last pair `rotate(0deg)` to `rotate(360deg)` would not be interpolated.  They would get collapsed "into" the second pair due to the mismatch.  I just want to check that asymmetry is the expected behaviour.</strike>

<strike>(Thanks for checking!  Certainly I agree the patch needs more work ... as noted in #619 I still find strange behaviour with an initial matrix animation.)</strike>

joeytwiddle:
<strike>But then there is a remaining issue: currently the fall back is also used when the two components match but are both _matrices_.  In that situation, shouldn't we interpolate the matrices, then move on to the next pair?</strike>

<strike>In other words, I believe the following shouldn't fall back.  But currently it does.</strike>

```
dt.play(new Animation(divs[0], [
  {transform: 'matrix(1,0,0,1,0,0)  rotate(0deg)  '},
  {transform: 'matrix(0,1,-1,0,0,0) rotate(360deg)'},
], timing));
```

<strike>I see you have already answered in agreement: _"I think that this patch has the correct behavior for the example you've provided"_</strike>&lt;/verbose&gt;

So I will try to fix this in a different branch...!

PR #620, opened 2014-07-03
[RFC] Interpolate paired matrices, avoid decomposing rest of chain

joeytwiddle:
This is the update for https://github.com/web-animations/web-animations-js/pull/618 since shans pointed out the issues with the former.

Prior to the patch, none of the following rotations were visible.  (Because the rotations were decomposed, which makes the start and end angles, 0 and 360, look the same, which makes interpolation go nowhere!)

With the patch applied, all of the following rotations become visible.  (In Chrome 34, after the patch, all the divs were rotating except for the third.)

```
dt.play(new Animation(divs[0], [
  {transform: "matrix(1,0,0,1,0,0) rotate(0deg)"},
  {transform: "matrix(1,0,0,1,200,0) rotate(360deg)"},
], timing));

dt.play(new Animation(divs[1], [
  {transform: "matrix(1,0,0,1,0,0)"},
  {transform: "matrix(1,0,0,1,100,0)"},
], timing));
dt.play(new Animation(divs[1], new KeyframeEffect([
  {transform: "rotate(0deg)"},
  {transform: "rotate(360deg)"},
], "add"), timing));

dt.play(new Animation(divs[2], [
  {transform: "scale(1) rotate(0deg)"},
  {transform: "matrix(1,0,0,1,200,0) rotate(360deg)"},
], timing));

dt.play(new Animation(divs[3], [
  {transform: "scale(1)"},
  {transform: "matrix(1,0,0,1,100,0)"},
], timing));
dt.play(new Animation(divs[3], new KeyframeEffect([
  {transform: "rotate(0deg)"},
  {transform: "rotate(360deg)"},
], "add"), timing));
```

I put caching in [another branch](https://github.com/joeytwiddle/web-animations-js/tree/decompose-matrices-individually-2-with-caching) because it wasn't very pretty.

### DIFF
--- a/test/testcases.js
+++ b/test/testcases.js
@@ -6,6 +6,7 @@ var tests = [
   'auto-test-color.html',
   'auto-test-color-names.html',
   'auto-test-composite-transforms.html',
+  'auto-test-composite-transforms-matrix-rotate.html',
   'auto-test-compositor.html',
   'auto-test-delay.html',
   'auto-test-element-animate.html',
--- a/test/testcases/auto-test-composite-transforms-matrix-rotate-checks.js
+++ b/test/testcases/auto-test-composite-transforms-matrix-rotate-checks.js
@@ -1,0 +1,46 @@
+timing_test(function() {
+  at(0 * 1000, function() {
+    assert_styles(".anim",{'transform':'matrix(1, 0, 0, 1, 0, 0)'});
+  });
+  at(0.4 * 1000, function() {
+    assert_styles(".anim", [
+      {'transform':'matrix(0.3090, 0.9511, -0.9511, 0.3090, 40, 0)'},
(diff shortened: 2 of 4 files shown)


joeytwiddle:
After discussion in [#629](https://github.com/web-animations/web-animations-js/issues/629#issuecomment-50847585) I realised that Chromium 34 was behaving differently from Chrome 36.  Specifically, `matrix` and `rotate` were behaving differently in Chrome 34, but behave the same in Chrome 36.  It may have been prefix-related.

Now that I can see `matrix()` and `rotate()` performing equivalently, I suspect this patch does not bring us closer to the spec.  (Nevermind that I _wish_ the spec would allow these rotations to interpolate!)  So this PR can probably be closed.

The testcase here is in `test/testcases/auto-test-composite-transforms-matrix-rotate.html`

PR #621, opened 2014-07-17
iterationComposite support for Motion Path

ewilligers:
Motion Path animations can now be constructed with an iterationComposite
setting of 'accumulate' to force the path translation to accumulate from
one iteration to the next.

### DIFF
--- a/web-animations.js
+++ b/web-animations.js
@@ -1882,11 +1882,25 @@ var clamp = function(x, min, max) {
 
 /** @constructor */
 var MotionPathEffect = function(path, autoRotate, angle, composite) {
+  var iterationComposite = undefined;
+  var options = autoRotate;
+  if (typeof options == 'string' || options instanceof String ||
+      angle || composite) {
+    // FIXME: add deprecation warning - please pass an options dictionary to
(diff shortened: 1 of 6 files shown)


alancutter:
We should have a spec link in the description for this.
http://www.w3.org/TR/web-animations/#the-motionpatheffect-interface

web-animations.js
```diff
@@ -1918,12 +1932,30 @@ MotionPathEffect.prototype = createObject(AnimationEffect.prototype, {
...
+    enterModifyCurrentAnimationState();
+    try {
+      // Use the default value if an invalid string is specified.
+      this._iterationComposite =
+          value === 'accumulate' ? 'accumulate' : 'replace';
+      this._updateOffsetPerIteration();
+    } finally {
+      exitModifyCurrentAnimationState(repeatLastTick);
+    }
+  },
```
alancutter, line 1948: No need to expose getter and setters for this.

alancutter, line 1948: I'm wrong, all AnimationEffects have this property: http://www.w3.org/TR/web-animations/#idl-def-AnimationEffect

PR #631, opened 2014-08-04
Add SVG Path interpolation

alancutter:
This is an upstream of steveblock's path interpolation code from the "countries" demo.


alancutter:
Fixed lint error, Travis should be happy now.

PR #645, opened 2014-10-21
Added strokeDashoffset and strokeWidth as properties

keanulee:


web-animations.js
```diff
@@ -4527,6 +4527,8 @@ var propertyTypes = {
   right: percentLengthAutoType,
   stopColor: colorType,
   stroke: colorType,
+  strokeDashoffset: lengthType,
+  strokeWidth: lengthType,
```
shans, line 4531: These need to be added to the svgProperties dictionary as well, otherwise they won't apply to SVG content.

shans:
Please provide a test of the new functionality.

